PR #17, opened 2015-02-05
fixed potential memory leaks

thrimbor:
There were some cases where memory could be leaked when realloc returned NULL. This should hopefully fix that.


dkl:
Hi, I think the idea is ok, but the free() calls should be simplified:

    if (oldResult != NULL)
        free( (void*) oldResult );
    =>
    free(oldResult);

thrimbor:
Done.

PR #19, opened 2015-07-04
properly implemented fb_hGetExePath for FreeBSD

thrimbor:
After stumbling across this issue (as documented on the freebasic.net-forum), I took the liberty of rewriting the FreeBSD-implementation of fb_hGetExePath, a part of the code is based on the linux version.
I tested it a few times, and, as far as I can tell, it does seem to do the job.


dkl:
Hi,

looks very nice. I'm going to add a changelog entry and merge it. Thanks.

PR #20, opened 2015-07-15
Lots of rtlib & gfxlib stuff for unix-like systems

thrimbor:
There are lots of small changes, so I'll provide a detailed list. With these changes, it's possible to run graphical fb-programs (with minor headaches remaining) on FreeBSD and OS X (and I think Solaris and NetBSD too, but I'm not sure right now).
If something's problematic (especially the makefile changes), please provide feedback so I can change the stuff.

Commit | Explanation | Tested on
--------- | --------------- | ------------
27bf4b6 | The NetBSD-specific code included a header that didn't exist in the same directory. Since it wasn't needed anyway, I simply removed the #include. | NetBSD 6.1.5 i386
e42f942 | NetBSD has several separate include-directories. I added the one for packages installed via "pkg" and the one for X11. I'm not sure this is enough though, because the X11 include path is version specific and could change in the future. | NetBSD 6.1.5 i386
5edcb59 | There wasn't a correct implementation of fb_hGetExePath for NetBSD yet, so I implemented it using /proc/curproc/exe and parts of the Linux-specific implementation. | NetBSD 6.1.5 i386
f90f829 | This patch isn't really needed because the old code worked fine already, but it simplifies the code a bit and reduces the room for problems. | Kubuntu 15.04 i386 with Linux 3.19.0-22
cd7d9ad | The fbc-makefile didn't find the headers for X11 and libffi on OS X because they're in different paths than on Linux. With this patch the makefile appends the X11- and libffi-specific search paths to the C-flags when building on OS X. They're not version-specific like the NetBSD ones, so they should be fine. | OS X Yosemite x86_64
f5e5879 | There was no implementation for fb_hGetExePath for OS X before, which lead to problems when trying to use exepath in FB-code or even trying to compile fbc. The code uses _NSGetExecutablePath and removes the "/." which OS X appends. | OS X Yosemite x86_64
a3e6c4d | hInit() was passing a const char * to tgestr which made the compiler complain. Since afaik ncurses doesn't change the passed string, I added a cast to char * which makes the complaint go away. | OS X, Kubuntu, NetBSD
d7a5306 | The linker on OS X likes to remove symbols when it thinks they aren't used. Because the __fb_errmsg-string wasn't initialized, it got removed on OS X which lead to problems when linking FB-programs. Now it's initialized to an empty string which fixes the problem (alternative would be to run ranlib -c on the rtlib). | OS X Yosemite x86_64
7bc051f | fb_hGetExeName wasn't implemented for OS X, but it's needed for running gfxlib2-programs, so I wrote an implementation based on the Linux-specific code and _NSGetExecutablePath. | OS X Yosemite x86_64
9ec36c6 | When not including X11-support on OS X, gcc couldn't resolve the ssize_t-type in gfx_x11.c. I fixed it by simply including sys/types.h. | OS X Yosemite x86_64
5f8adbf | gfxlib2 wasn't allowing to use X11 on OS X. I added the option to enable X11 support by passing XQUARTZ_ENABLE=1 to the makefile. This allows to run gfxlib2-programs on OS X. However, for this to work, the installation of XQuartz is required. Since XQuartz isn't installed by default on fresh OS X installations, the XQuartz support is disabled by default. I consider this a temporary workaround until gfxlib2 gets a native Cocoa-backend. | OS X Yosemite x86_64
75ef944 | The rtlib didn't have an implementation of fb_hGetExePath for Solaris yet, so I added one using the Solaris-specific procfs. | OpenIndiana (Illumos) 5.11 i386
6f21bbb | Implementation of fb_hGetExeName for Solaris. Similar to the above commit. | OpenIndiana (Illumos) 5.11 i386
4e93bea | The makefile wasn't able to detect Solaris/SunOS automatically by using uname, so I fixed that. | OpenIndiana (Illumos) 5.11 i386
29df4d2 | For some reason, gcc on Solaris complained about the usage of alloca (something about redefinition...). After fiddling around for a while I resolved this by using __builtin_alloca instead, which seems to work. I'm not sure if this is a permanent fix (or even a good one). | OpenIndiana (Illumos) 5.11 i386
8efed13 | Implementation of fb_hGetExeName for FreeBSD. Similar to the fb_hGetExePath implementation, also uses the KERN_PROC_PATHNAME syscall. | FreeBSD 10.1-RELEASE i386
387cc47 | Implementation of fb_hGetExeName for NetBSD, similar to the fb_hGetExePath implementation. Also uses /proc/curproc/exe (NetBSD-specific procfs). | NetBSD 6.1.5 i386

Some notes:
- The OS X implementations of fb_hGetExePath and fb_hGetExecutablePath use _NSGetExecutablePath. As far as I remember, the returned path can contain symlinks, so maybe using realpath() could be a good idea. This should probably be investigated and adressed in a future commit.
- Running fbc on OS X works fine now. Unfortunately, the out-of-the-box experience on OS X still isn't very good, because "as" on OS X has problems with the flags fbc passes. See this pastebin-link for details: http://pastebin.com/emE3iF0T
- When not explicitly passing "-asm att" on OS X, gcc complains: "qbsprite.c:1:0: error: -masm=intel not supported in this configuration"
- As i wrote above, using XQuartz on OS X is not a perfect solution. I played around with Cocoa a bit, and maybe I'm able to come up with a rough prototype of a Cocoa-backend in the near future. But I can't promise anything.
- Talking about gfxlib-backends, a Wayland-backend would be nice too, although it's not really necessary right now since X11 will remain available on Linux systems in the near future.
- I wanted to implement fb_hGetExePath for OpenBSD too, but it seems that OpenBSD doesn't provide an interface for that. There is no procfs on OpenBSD by default, and it also doesn't provide a FreeBSD-like syscall for that. I already don't like this system.
- Some platforms (Solaris for example) use rather old versions of gcc (4.4 I think). These old versions seem to have problems with the C-code fbc emits. This could be a problem when there's no asm-backend available.
- Afaik the fb_hGetExeName function isn't directly available to FB-programs right now. It might be a good idea to add that, since it can differ from what command(0) tells.

And here's proof of gfxlib2 on OS X ;)
[gfxlib on OS X](http://www.directupload.net/file/d/4049/ljbjjdzx_png.htm)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -23,6 +23,7 @@ Version 1.04.0
 - On Darwin, linking with rtlib failed due to __fb_errmsg being discarded
 - Correctly implemented fb_hGetExeName for FreeBSD, NetBSD, Darwin and Solaris (needed by gfxlib2)
 - The FB makefile can now detect the OS when building on Solaris
+- fbc will now pass the proper parameters to "as" on OS X
 
 
 Version 1.03.0
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,8 @@ Version 1.04.0
 [changed]
 
 [added]
-- exepath() now works on FreeBSD
+- exepath() now works on FreeBSD, NetBSD, Darwin and Solaris
+- The FB gfxlib can now be built on OS X using XQuartz by passing ENABLE_XQUARTZ=1 to make
 
 [fixed]
 - False-positive "ambigious sizeof" warnings if the identifier could refer only to a type or procedure (in that case it should already be fairly clear that the type will be chosen, not the procedure, because sizeof() can't be applied to procedures at all), or if an expression such as "array(0)" is given to sizeof() (i.e. something that starts with an identifier but is followed by further tokens that clarify that it's not refering to a type)
@@ -18,6 +19,10 @@ Version 1.04.0
 - __FB_ASM__ wasn't #defined for x86_64 even though the -asm option can be used on x86_64 too
 - C backend: UDTs using Field=N were emitted incorrectly (causing gcc errors), if they contained a field of type of another UDT that didn't use Field=N.
 - False-positive "ambigious sizeof" warnings if the identifier could refer to a type or variable of that type (in this case the size is obviously the same, and it doesn't matter which one is used)
+- rtlib & gfxlib compilation was broken on NetBSD, Darwin and Solaris (header & include-path issues)
+- On Darwin, linking with rtlib failed due to __fb_errmsg being discarded
+- Correctly implemented fb_hGetExeName for FreeBSD, NetBSD, Darwin and Solaris (needed by gfxlib2)
+- The FB makefile can now detect the OS when building on Solaris
 
 
 Version 1.03.0
--- a/src/compiler/fbc.bas
+++ b/src/compiler/fbc.bas
@@ -772,15 +772,17 @@ private function hLinkFiles( ) as integer
 			end if
 		end if
 
-		'' All have crti.o, except OpenBSD
-		if (fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_OPENBSD) then
-			ldcline += hFindLib( "crti.o" )
-		end if
+		if (fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_DARWIN) then
+			'' All have crti.o, except OpenBSD and Darwin
+			if (fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_OPENBSD) then
+				ldcline += hFindLib( "crti.o" )
+			end if
 
-		if( fbGetOption( FB_COMPOPT_PIC ) ) then
-			ldcline += hFindLib( "crtbeginS.o" )
-		else
-			ldcline += hFindLib( "crtbegin.o" )
+			if( fbGetOption( FB_COMPOPT_PIC ) ) then
+				ldcline += hFindLib( "crtbeginS.o" )
+			else
+				ldcline += hFindLib( "crtbegin.o" )
+			end if
 		end if
 
 	case FB_COMPTARGET_XBOX
@@ -810,7 +812,9 @@ private function hLinkFiles( ) as integer
 	'' Begin of lib group
 	'' All libraries are passed inside -( -) so we don't need to worry as
 	'' much about their order and/or listing them repeatedly.
-	ldcline += " ""-("""
+	if ( fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_DARWIN ) then
+		ldcline += " ""-("""
+	end if
 
 	'' Add libraries passed by file name
 	scope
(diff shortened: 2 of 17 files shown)


dkl:
Hi,

most of this looks good. It's nice to see someone working on these FB ports.

Could you add changelog.txt entries, for user-visible changes? (usually the way we do it, if a commit fixes a bug or adds a feature etc., it also adds an entry to changelog.txt) Even if it's a build fix, I think it should be noted in the changelog, for example "The FB gfxlib can now be built on MacOSX using XQuartz".

I agree it would be very nice to have Cocoa and Wayland versions of FB's gfxlib. Cocoa is needed for a full MacOSX port, and Wayland is nice to have when it starts replacing X11 on GNU/Linux. Another idea was to make an SDL port, since that would make the gfxlib work on many more systems too.

If "as" on MacOSX doesn't accept the --32/--64 options, no problem, we can adjust fbc to not pass them for the target. Maybe there are alternative options to use? Or does the assembler only support one target anyways? On GNU/Linux systems it's common to have binutils supporting both 32bit and 64bit, for example.

If -masm=intel is totally unsupported, maybe we should make -asm att the default then for the target. It's a bit unfortunate of course since -asm intel is the FB default in general.

I know of at least one problem with the C backend, that means it doesn't yet work with older gcc's: sometimes there can be duplicate typedefs emitted, which older gcc's didn't allow.

thrimbor:
Hi,

I added the changelog.txt entries, I hope they're okay like that.

"as" seems to accept different parameters on OS X. Instead of "--32" fbc needs to pass "-arch i486", and "-arch x86_64" instead of "--64" ("as" even offers arm, ppc, sparc, m68k...). I added another commit that fixes this problem. Also, "--strip-local-absolute" isn't added on OS X anymore because it lead to an error. The commit also fixes some other small stuff in fbc, like not adding the X11 libraries on OS X etc.
With this additional commit, fbc is now able to compile and link on OS X. It needs "-asm att" and throws a few warnings, but it works.
I was even able to build fbc and the libs on OS X with only make ENABLE_XQUARTZ=1 FBFLAGS="-asm att".

I also thought about starting and SDL-backend. However, that would mean that SDL needs to be installed in order to run FB programs. That may be a good option for porting FB to new systems, but I'm not sure whether this is a nice solution.

>I know of at least one problem with the C backend, that means it doesn't yet work with older gcc's: sometimes there can be duplicate typedefs emitted, which older gcc's didn't allow.

I think that this was the problem. I may be able to check when my exams are over.

changelog.txt
```diff
@@ -18,6 +19,12 @@ Version 1.04.0
 - __FB_ASM__ wasn't #defined for x86_64 even though the -asm option can be used on x86_64 too
 - C backend: UDTs using Field=N were emitted incorrectly (causing gcc errors), if they contained a field of type of another UDT that didn't use Field=N.
 - False-positive "ambigious sizeof" warnings if the identifier could refer to a type or variable of that type (in this case the size is obviously the same, and it doesn't matter which one is used)
+- rtlib & gfxlib compilation was broken on NetBSD, Darwin and Solaris (header & include-path issues)
+- rtlib was ignoring const-correctness in hInit when calling tgetstr
+- Darwins linker was removing the __fb_errmsg symbol in rtlib
+- Correctly implemented fb_hGetExeName for FreeBSD, NetBSD, Darwin and Solaris (needed by gfxlib2)
+- Makefile wasn't able to detect the OS when building on Solaris
+- fbc was passing parameters to "as" that weren't valid on OS X
```
dkl: I think some of these should be rewritten, maybe like this:

- rtlib & gfxlib compilation was broken on NetBSD, Darwin and Solaris (header & include-path issues)
- On Darwin, linking with rtlib failed due to __fb_errmsg being discarded
- Correctly implemented fb_hGetExeName for FreeBSD, NetBSD, Darwin and Solaris (needed by gfxlib2)
- The FB makefile can now detect the OS when building on Solaris
- fbc will now pass the proper parameters to "as" on OS X

src/compiler/fbc.bas
```diff
@@ -3097,12 +3113,18 @@ private sub hAddDefaultLibs( )
...
+			#if defined(__FB_DARWIN__) and defined(ENABLE_XQUARTZ)
+				fbcAddDefLibPAth( "/opt/X11/lib" )
+			#endif
 
-			fbcAddDefLib( "X11" )
-			fbcAddDefLib( "Xext" )
-			fbcAddDefLib( "Xpm" )
-			fbcAddDefLib( "Xrandr" )
-			fbcAddDefLib( "Xrender" )
+			#if not(defined(__FB_DARWIN__)) or defined(ENABLE_XQUARTZ)
```
dkl: Here the `not` operator should probably be put inside the parentheses:
`#if (not defined(__FB_DARWIN__)) or defined(ENABLE_XQUARTZ)`
so it's clear that it should affect only the one `defined()`, not the entire `or` operation. I'm never sure about `not` operator precedence...

thrimbor:
Ok, I changed the commits.

dkl:
Allright, thanks very much.

PR #24, opened 2015-10-13
Fix FileAttr return value type for x86_64 on Windows

swurzinger:
rtlib: The fb_FileAttr function stores its return value (type ssize_t) internally in a variable "ret" of type int. This causes problems on Windows x86_64, because int will be 32-bit and ssize_t 64-bit. This fix changes the internal "ret" variable to type ssize_t to match the return type.


dkl:
Oh, looks like this was missed in commit a86c62a199ebf91d1870ce3e65082249e0a12002. Thanks.

PR #25, opened 2015-10-25
fixed linker warnings on OS X

thrimbor:
Just a quick adjustment to the parameters that fbc passes to ld. Similar to the changes in the emscripten branch, but I also added "-macosx_version_min 10.6" because it's expected on OS X.
fbc requires at least OS X 10.6 Snow Leopard, but since 10.6 was assumed by default the requirements don't change.

Tested on OS X 10.10.5 Yosemite.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -10,6 +10,7 @@ Version 1.05.0
 - Type<Foo>(...) expressions couldn't be used at the begin of a statement (because the Type keyword was treated as begin of a Type declaration)
 - FileAttr() was still broken for 64bit - the result value was truncated to 32bit
 - 1.04.0 regression: Under -gen gcc -asm att, support for gcc-style inline asm was broken
+- ld was warning about unsupported linker options on OS X
 
 
 Version 1.04.0
--- a/src/compiler/fbc.bas
+++ b/src/compiler/fbc.bas
@@ -664,8 +664,10 @@ private function hLinkFiles( ) as integer
 		'' (needed until binutils' default DJGPP ldscripts are fixed)
 		ldcline += " -T """ + fbc.libpath + (FB_HOST_PATHDIV + "i386go32.x""")
 	else
-		'' Supplementary ld script to drop the fbctinf objinfo section
-		ldcline += " """ + fbc.libpath + (FB_HOST_PATHDIV + "fbextra.x""")
+		if( fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_DARWIN ) then
+			'' Supplementary ld script to drop the fbctinf objinfo section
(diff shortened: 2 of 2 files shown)


rversteegen:
May I ask why you picked 10.6 as the requirement? Have you tried to run FB programs on 10.5 or earlier? At least using the old Mac port of FB 0.22, fbc and fb programs ran fine on OS 10.5 when specifying -macosx_version_min 10.4. Unfortunately I no longer have access to a machine older than 10.7.5 to test on, but I know someone with 10.5 who I'll ask.

If there's no good reason otherwise, I think -macosx_version_min should be 10.4 (the oldest version supporting x86), which works fine for me on 10.8.5. It apparently does not require using a matching OSX 10.4u SDK. [This link](http://stackoverflow.com/questions/2923502/what-does-macosx-version-min-imply) describes a few things these flags do. System header files should be translated to .bi matchubg -macosx_version_min. When targeting later OSX versions a number of standard functions like fwrite get swapped out for alternative versions with suffixes appended to their names, such as fwrite$UNIX03.

thrimbor:
Unfortunately, the only version I had available at this time was OS X 10.10, so I wasn't able to test with earlier versions. I chose 10.6 because it was already assumed on Yosemite when not specifying -macosx_version_min.
It shouldn't be a problem to change this later (I assume dkl would happily accept a patch), once somebody verified that fbc indeed works on that older version. Maybe we can go back even further if fbc works on the old PowerPC-versions of OS X.

rversteegen:
Well, I have been translating some system headers, and am sticking to 10.4. Trivial to meet this requirement in the rtlib, maybe not in a hypothetical Cocoa backend for fbgfx.

I don't know what happens if you pass -macosx_version_min (called the "deployment target") to the linker and try to link with some library that requires something higher. Possibly not good. Perhaps we need to add a command line option to fbc to specify the deployment target. That's what [FPC did](http://wiki.freepascal.org/FPC_New_Features_2.6.2#Darwin.2FMac_OS_X). In older versions of FPC you had to [pass options to the linker manually](http://wiki.lazarus.freepascal.org/OS_X_Programming_Tips#Deploying_an_executable.2C_Compiling_under_10.6_for_older_version_of_Mac_OS_X).

Looking further into the use of -macosx_version_min X and compiling against the OSX SDK version Y (where X <= Y) I've read that such a binary will run on all versions >= X, however if you use an API introduced after X but present in Y, the compiler will (without warning) produce a weakly linked symbol which is loaded as NULL at run time if missing. You are responsible for testing if the address is NULL at runtime. So compiling against an older SDK isn't necessary, but if you're using lots of OSX APIs it's the [only practical way](https://openradar.appspot.com/radar?id=1850406) to avoid accidentally breaking support for an older version.

PR #27, opened 2016-04-25
LLVM backend fixes

thrimbor:
I saw that the LLVM-backend didn't work on my Arch Linux machine and decided to fix that.

Commit fca08f4 addresses the problem that llc doesn't emit ".intel_syntax noprefix" which makes gas fail. As explained in the commit message, the emitted assembly syntax doesn't really matter anyway (inline assembly is transformed by llc), so I just let llc use the AT&T syntax. I do consider this a workaround for a llc-bug, but it's doesn't hurt and does get things working.

Commit 2f0680b changes the LLVM-IR-emitter to use the new load-syntax, which was (afaik) introduced with LLVM 3.7. Since this version was released last August (and LLVM 3.8 has been released already, too), I think we shouldn't have any problems with assuming that LLVM 3.7+ is available. Well, except for Debian, but I don't think there's anything we can do about it: We either break Debian, or everything else.
This commit is the first time I actually touched any compiler internals, so please fell free to tell me if there's anything problematic in there. But since it's a relatively small change I think it's fine.

With these two changes, the LLVM-backend now works on my Arch Linux x86_64 system, but I haven't been able to test it on any other platforms yet.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -15,6 +15,7 @@ Version 1.06.0
 - #797: False-positive warnings when mixing booleans with comparisons in logic operations
 - TRE binding: Didn't compile with TRE_USE_SYSTEM_REGEX_H
 - LLVM backend: Let llc emit AT&T syntax assembly to prevent gas from failing
+- LLVM backend: Emit the new (LLVM 3.7+) load-syntax
 
 
 Version 1.05.0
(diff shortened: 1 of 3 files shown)


dkl:
Hello,

The load fix looks good, thanks. I've cherry-picked it as 954b41a581cc95990ffd616d279072d4cbbbd4e4. Looks like LLVM IR will see further changes in this area in the future (removal of typed pointers).

About the x86 Intel syntax issue - LLVM 3.8 now emits the `.intel_syntax` directive:
https://llvm.org/bugs/show_bug.cgi?id=23391
So it works now. And for older llc versions, it's possible to use `fbc -gen llvm -asm att` as a work-around. I think we should keep -asm intel/att support as with other backends.

PR #28, opened 2016-04-30
Fix UTF to wchar conversion functions

rversteegen:
I noticed that recently added UTF conversion testcases were failing; the newly rewritten conversion code was still not right.

There are completely different codepaths depending on the size of wchar, and unfortunately I don't see any simple way to run testcases against different wchar sizes. The simplest would probably be to expose all the currently private functions so that they can be accessed by tests. I tested on a 32 bit wchar machine.

### DIFF
--- a/src/rtlib/utf_convto_wchar.c
+++ b/src/rtlib/utf_convto_wchar.c
@@ -241,6 +241,27 @@ static FB_WCHAR *hUTF8ToWChar( const UTF_8 *src, FB_WCHAR *dst, ssize_t *chars )
 	return res;
 }
 
+static FB_WCHAR *hUTF16ToUTF16( const UTF_16 *src, FB_WCHAR *dst, ssize_t *chars )
+{
+	/* Have to determine and return actual string length */
+	ssize_t len = 0;
+	if( dst == NULL ) {
+		while( src[len] )
+			len++;
+		dst = malloc( (len + 1) * sizeof( UTF_16 ) );
+		memcpy( dst, src, (len + 1) * sizeof( UTF_16 ) );
+	} else {
+		while( src[len] && len < *chars )
+			len++;
+		memcpy( dst, src, len * sizeof( UTF_16 ) );
+		if (len < *chars)
+			/* The input buffer has a trailing NUL character, copy it */
+			dst[len] = 0;
+	}
+	*chars = len;
+	return dst;
+}
+
 static FB_WCHAR *hUTF16ToUTF32( const UTF_16 *src, FB_WCHAR *dst, ssize_t *chars )
 {
 	UTF_16 c;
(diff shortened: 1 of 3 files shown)


src/rtlib/utf_convto_wchar.c
```diff
@@ -241,6 +241,26 @@ static FB_WCHAR *hUTF8ToWChar( const UTF_8 *src, FB_WCHAR *dst, ssize_t *chars )
...
+	} else {
+		while( src[len] && len < *chars )
+			len++;
+		memcpy( dst, src, len * sizeof( UTF_16 ) );
+		if (len < *chars)
+			/* The input buffer has a trailing NUL character, copy it */
+			dst[len] = 0;
+	}
+	*chars = len;
+}
```
dkl: missing `return dst;` here (found by compiler warning)

src/rtlib/utf_convto_wchar.c
```diff
@@ -313,13 +333,6 @@ static FB_WCHAR *hUTF16ToWChar( const UTF_16 *src, FB_WCHAR *dst, ssize_t *chars
...
 
 	case sizeof( UTF_16 ):
-		if( dst == NULL ) {
-			dst = malloc( (*chars + 1) * sizeof( UTF_16 ) );
-			memcpy( dst, src, *chars * sizeof( UTF_16 ) );
-			dst[*chars] = 0;
-		} else {
-			memcpy( dst, src, *chars * sizeof( UTF_16 ) );
-		}
 		break;
```
dkl: missing call to `hUTF16ToUTF16()` here, `dst = hUTF16ToUTF16( src, dst, chars );` (found by compiler warning)

dkl:
Hello,

looks like the problem is that the `memcpy` cases of `UtfToWchar` didn't handle the `chars` parameter like the other cases. The fix makes sense to me.

This just made be realize again how broken these functions seem to me.
1. `UtfTo[W]Char` expects its UTF input to be null-terminated, and `[W]CharToUtf` doesn't. Shouldn't it be the other way round? FB's Wstrings/Zstrings (which is what Wchar/Char refers to here) are null-terminated, but a buffer containing some UTF data shouldn't have to be.
2. `UtfTo[W]Char` with dst=NULL ignores the chars value (and just reads until a null char), unlike the pre-allocated dst code paths. That's totally inconsistent.

rversteegen:
Damn, that show how bad untestable code is... and how dangerous last minute changes are (I only checked that the tests still passed (since only 16 bit wchar was broken) and missed the compiler warnings.

The semantics that these functions implement is indeed strange. That's why I had to very carefully document it before fixing those two cases to match the others. (It's too bad that the rtlib is so poorly commented.)

But maybe the semantics are too much trouble to change if the callers are few, and always respect them. fb_UTFToWChar is not called anywhere in the rtlib. fb_WCharToUTF is called from one place in the rtlib.  AFAICT none of the utf to/from wchar/char functions are generated by fbc. I also note that fbc reimplements hUTF8ToChar in lex-utf.bas.

Wow, I just noticed that hUTF8ToUTF16 and other functions are duplicated in rtlib/dev_file_encod_read_core.c! This is a mess. Sorry, I can't say I'm interested in spending further time on this. I only investigated this bug because I mistakenly assumed that the failing testcase was a bug in the darwin port.

dkl:
The functions are accessible via `utf_conv.bi` (as opposed to being built-in keywords), though I don't think they're widely used.

PR #29, opened 2016-05-11
Follow linux/darwin 16-byte stack alignment ABI rule; allow linking C/C++ into tests

rversteegen:
This fixes [bug #659](https://sourceforge.net/p/fbc/bugs/659/).

Apologies for mixing two seemingly unrelated things, but to write proper testcases for following a platform's ABI I considered it sensible to actually link to C. More such ABI testcases may be sensible in future. And I discovered an orphaned testcase in the process; turns out 'extern "C++"' is broken in 64 bit builds.


dkl:
Hello,

this looks very interesting.

However, when I built the test suite with the new compiler, it crashed (linux-x86) in one of its initialization functions (a sub without parameters). So I tested this code:
```
sub f()
end sub

f()
```
and it crashes too. Apparently there is a problem with the code generation changes.

Here's the diff between assembly code generated by old and new compiler for linux-x86:

```diff
--- 1-old.asm	2016-05-11 18:55:40.591747637 +0200
+++ 1-new.asm	2016-05-11 18:55:27.996161058 +0200
@@ -6,6 +6,9 @@
 .globl F
 F:
 .type F, @function
+push ebp
+mov ebp, esp
+sub esp, 8
 .Lt_0004:
 .Lt_0005:
 ret
@@ -15,22 +18,25 @@
 .globl main
 main:
 .type main, @function
-push ebp
-mov ebp, esp
 and esp, 0xFFFFFFF0
 sub esp, 4
+push ebp
+mov ebp, esp
+sub esp, 8
 mov dword ptr [ebp-4], 0
+sub esp, 4
 push 0
 push dword ptr [ebp+12]
 push dword ptr [ebp+8]
 call fb_Init
-add esp, 12
+add esp, 16
 .Lt_0002:
 call F
 .Lt_0003:
+sub esp, 12
 push 0
 call fb_End
-add esp, 4
+add esp, 16
 mov eax, dword ptr [ebp-4]
 mov esp, ebp
 pop ebp
```

I suppose the problem are the added ```sub esp, 8``` instructions which aren't counteracted or maybe even shouldn't be added in the first place (I'm not sure).

rversteegen:
I see: I did not update hDestroyFrame, as I forgot that it skips "pop ebp" if it finds that it can get away without a stack frame. However, I was compiling the cunit-tests with -g, which forces it to always push ebp. What's crazy is that all the log-tests pass even without -g, so I didn't notice this. I'll push a fix when I have more time tonight.

In fact, as you notice from the diff, it aligns the stack even for an empty sub, and anything with locals requires a frame pointer (mv esp, ebp), so I think that alignment completely eliminates stackframeless functions.

dkl:
Thanks. It works now, the tests run fine on linux-x86.

For win32, I had to add `CC := gcc` (which we also have in the toplevel makefile) to tests/bmk-make.mk, otherwise the new `stack-alignment.bmk` complained about `cc` (`CC` default value) not being found. I suppose MinGW-w64 doesn't provide a "cc".

dkl:
There's another problem: parameters of main() aren't accessed properly anymore, because esp is changed before being stored into ebp. Example for testing:

```
print __FB_ARGC__
```

rversteegen:
Oops. Here's a fix. Only tested on Linux (the new testcase has platform-variable behaviour).

rversteegen:
Apparently github ignores further commits pushed to a branch after it is merged (makes sense), but you can get the fix by merging my stackalign branch:
https://github.com/rversteegen/fbc/commit/f893281f5d64511fcc764cccf7fb6df21ec0d268

dkl:
Thanks!

PR #38, opened 2017-08-31
832Fix - Fix bug allowing QB style suffixes on all keywords

ImortisInglorian:
832Fix - Fix bug allowing QB style suffixes on all keywords, regardless of -lang

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -41,6 +41,7 @@ Version 1.06.0
 - #846: Fix compiler crash on global variable initializer with type<T>(...) expression where T isn't a UDT
 - #847: Windows API binding: Fixed winnt.bi SECURITY_*_AUTHORITY initializers
 - #851: '#LINE line filename' only changed the source filename in error messages, not in debug info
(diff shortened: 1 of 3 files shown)


dkl:
It seems good to me to disallow the suffixes in the lexer's identifier parsing.

There is a way to check whether the current dialect allows suffixes, which I think should be used here:
```
if( fbLangOptIsSet( FB_LANG_OPT_SUFFIX ) ) then ...
```
`FB_LANG_OPT_SUFFIX` is enabled in fb.bas `langTb()` array for qb, deprecated and fblite.

ImortisInglorian:
I will make the suggested changes.  Thanks.

ImortisInglorian:
Changes made.

dkl:
Oops, that should have been `if( fbLangOptIsSet( FB_LANG_OPT_SUFFIX ) = FALSE ) then ...`, otherwise the logic is negated.

ImortisInglorian:
Whoops.  I didn't even think to look.  One more time.

PR #40, opened 2017-09-15
oGLfbGFX - Added full openGL 2D render for fbGFX

ImortisInglorian:
See forum post: http://www.freebasic.net/forum/viewtopic.php?f=17&t=25902

To enable 2d rendering when Flip is called:
screencontrol 150,1

To enable automatic 2d rendering:
screencontrol 150,2

To set a zoom factor of 2:
screencontrol 151,2

Then use the openGL mode on the screen command you your choice.

Examples will be included if this gets excepted into the base.


dkl:
In general I like the idea, and it should be safe too, since as far as I can see, the non-OpenGL graphics mode is not affected.

PR #42, opened 2017-09-30
Windows lib/header patch for FB1.05

stephanbrunker:
from Forum post: [https://www.freebasic.net/forum/viewtopic.php?f=14&t=25888](url)
I added a new folder named "patch" and put there the changed files: 

Changed definition files:
def\comctl32.def: added LoadIconMetric
def\user32.def: added CalculatePopupWindowPosition

Changed Headers:
inc\win\commctrl.bi: changed _WIN32_WINNT value for LoadIconMetric
inc\win\shellapi.bi: changed ShellNotifyIcon

Compiled new import libraries with from the windows libraries with these definition files:
Lib\libcomctl32.dll.a
Lib\libuser32.dll.a

The headers and libraries are in the patch.zip file, users of FB v1.05 can now patch their libraries with these files fo fix the issues above.


dkl:
I wish fbc could use the Windows API import libraries from the MinGW.org or mingw-w64 projects, to avoid duplicating the work. That's why I removed many .def files from fbc in the past, and I don't like the idea of re-adding them. If some functions are missing from some Windows API import libraries, in my opinion the changes should be submitted to the mingw-w64 project.

This pull request isn't ready to merge, because it adds a patch/ sub-directory instead of modifying the existing source code.

stephanbrunker:
Never mind. I linked the forum post about the missing functions to my fork, so the patch can be found. Don't know about any better way to host such a patch. Some of the missing functions are actually corrected on mingw-w64, but that doesn't help until the release of a new FB version. The other kind of issue is the automatic translation of the C headers which leads to incorrect win32_winnt values and unicode errors. The def-File is actually the same as in the mingw-w64 repository now or should be (the function vanished temporarly on the time of compiling FB1.05 because of some tidying) The real fix would be to fix that automatic translation which is really not easy, and for the quick fix now and the unexperienced user ... you know where to find it.

dkl:
I'm going to try and take another look at the binding generation to add more _WIN32_WINNT choices.

PR #44, opened 2017-10-04
Dynamic libraries in DOS

angros47:
I integrated in the last version some changes made in 2013 by Monochromator. This patch allows the creation and use of Dynamic Link Libraries in DOS, using dxe3gen (that is included in djgpp)

The file symb_reg.txt has been created using a tool called maksymbr by Monochromator.

### DIFF
--- a/makefile
+++ b/makefile
@@ -564,9 +564,6 @@ ifeq ($(TARGET_OS),dos)
   # Avoid hitting the command line length limit (the libfb.a ar command line
   # is very long...)
 	$(QUIET)rm -f $@
-	$(AR) x $(DJDIR)/lib/libpthread.a
-	$(AR) x $(DJDIR)/lib/libsocket.a
-	mv -f *.o $(libfbmtobjdir)
 	$(QUIET_AR)$(AR) rcs $@ $(libfbmtobjdir)/*.o
 else
 	$(QUIET_AR)rm -f $@; $(AR) rcs $@ $^
--- a/makefile
+++ b/makefile
@@ -260,7 +260,6 @@ endif
 ifeq ($(TARGET_OS),dos)
   FBNAME := freebas$(ENABLE_SUFFIX)
   FB_LDSCRIPT := i386go32.x
-  DISABLE_MT := YesPlease
 else
   FBNAME := freebasic$(ENABLE_SUFFIX)
   FB_LDSCRIPT := fbextra.x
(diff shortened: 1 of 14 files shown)


angros47:
It also allows the use of threads, using pthreads from pth207b.zip, and socket library from ls080b.zip. Since those libraries haven't been updated in a while, installing them in djgpp is a bit tricky: I first installed all the files in pth207b.zip, and then installed in the lib directory only the file libsocket.a from ls080b. I also had to edit the file pthread.h (in the include directory of djgpp installation), removing the lines:

#include <sys/socket.h> /* for sockaddr /
#include <sys/wtime.h> / for struct timespec */
#include <sys/select.h>

makefile
```diff
@@ -561,7 +560,17 @@ $(LIBFBPIC_C): $(libfbpicobjdir)/%.o: %.c $(LIBFB_H) | $(libfbpicobjdir)
 	$(QUIET_CC)$(CC) -fPIC $(ALLCFLAGS) -c $< -o $@
 
 $(libdir)/libfbmt.a: $(LIBFBMT_C) $(LIBFBMT_S) | $(libdir)
+ifeq ($(TARGET_OS),dos)
+  # Avoid hitting the command line length limit (the libfb.a ar command line
+  # is very long...)
+	$(QUIET)rm -f $@
+	$(AR) x $(DJDIR)/lib/libpthread.a
```
dkl: Is it necessary to include libpthread and libsocket into libfbmt? I expected additional dependencies to be added to the default libs in fbc.bas:hAddDefaultLibs() (i.e. -lpthread -lsocket would be added to fbc's link command).

src/rtlib/dos/symb_reg.txt
```diff
@@ -0,0 +1,1661 @@
+extern_asm(_fb_WstrInstrAny);
```
dkl: Can the tool to generate this file be included in the source tree, so others can use it to update this file (or is it not supposed to be changed)? Can this file be automatically generated by the makefile?

angros47: The tool consists into a .bat file, and an executable (with .bas source included), built by Monochromator. I guess it could be included, but comments and texts are in a language I can't identify. I had to use that tool to rebuild symb_reg.txt, since it had to be changed from Freebasic 0.90. I can provide you the files... not sure how to integrate it in the makefile

angros47: Actually, I checked, and it cannot be automatically generated by the makefile. In fact, the program extract the symbol names from the file libfb.dxe, that is built from libfb.a . Basically, it needs a copy of the runtime library that is already compiled, then it generates the updated symbol list (at this point, it's possible to replace the file symb_reg.txt with the updated one, delete the .o files and rebuild the library). So, it can't be automated in the makefile.

makefile
angros47: I followed the building procedure used by Monochromator. Perhaps your solution could work as well, I haven't tested it

src/rtlib/dos/symb_reg.txt
dkl: Ah, ok thanks. Could you please add the maksymbr.bas and bat script to src/rtlib/dos, and translate it to English if at all possible? It seems very important to have in the repository together with the feature, since I suppose it will have to be updated everytime the public interface of libfb is changed.

makefile
dkl: In hAddDefaultLibs(), it's possible to check for `fbGetOption( FB_COMPOPT_MULTITHREADED )`, and only add the pthread/socket libs in that case. That would prevent them from being linked automatically in the non-mt case.

src/rtlib/dos/symb_reg.txt
angros47: Translated.... the characters seemed to be Cyrillic, and it was in Russian. I used Google translator, for it

PR #47, opened 2017-10-22
fbdoc & tools update Oct 2017

jayrm:
Main changes are:
- copyright updates for 2017
- support for custom certificate file to work with https://www.freebasic.net/wiki/
- tools should recognize both http:// & https:// links in string comparisons
- wiki snapshot


doc/fbchkdoc/spell.bas
```diff
@@ -122,7 +124,7 @@ function GetWord( byref text as string, byref i as integer, byref j as integer )
 		wend
 
 		if( j <= len(text) ) then
-			if( lcase(mid( text, j, i - j + 1 )) = "http:" ) then
+			if( (lcase(mid( text, j, i - j + 1 )) = "http:") or (lcase(mid( text, j, i - j + 1 )) = "https:") ) then
```
dkl: Is the length `i - j + 1` also correct for "https:" here? There probably should be a `startsWith()` helper function for such cases.

jayrm:
dkl, thanks.
It's correct.  In this function, a "word" is any sequence of "word-characters" where a word-character is any number, letter, apostrophe, or underbar.  In the previous `while` block, `'j'` is starting on word-character, and `'i'` is scanned forward to next non-word-character, which includes colon `':'`.  If the word scanned so far starts with "http:" or "https:" then it scans the remainder of the "word" using characters found in urls.

```' EXAMPLE
dim as integer i = 1, j
dim as string s = "text1 http://www.freebasic.net text2 https://www.freebasic.net text3"
while i <= len(s)
	print GetWord( s, i, j )
wend

/' OUTPUT:
text1
http://www.freebasic.net
text2
https://www.freebasic.net
text3
'/
```

dkl:
Ah, ok then, that makes sense.

PR #49, opened 2017-10-23
Fbdoc fixes: wiki URL and HTML character escaping

swurzinger:
The commits in this PR aim to fix the character escaping issue discussed here:
https://freebasic.net/forum/viewtopic.php?f=9&t=25972

It allows the following HTML character escape sequences:
```
&gt;
&lt;
&amp;
&quot;
&#xxx; (where xxx represents a number between 0 and 255)
```

In tables the sequence `&#59;` (semicolon) is not allowed, because semicolon is used as a table cell separator.

Additionally it changes the Wiki URL to use https:// instead of http://
The issue is mentioned here:
https://freebasic.net/forum/viewtopic.php?p=235627#p235627

### DIFF
--- a/doc/fbchkdoc/fbchkdoc.org
+++ b/doc/fbchkdoc/fbchkdoc.org
@@ -16,4 +16,3 @@ dev_wiki_url =
 ' dev_certificate = path/filename.pem"
 dev_username = 
 dev_password = 
-
--- a/doc/libfbdoc/CWakka2Html.bas
+++ b/doc/libfbdoc/CWakka2Html.bas
@@ -758,7 +758,7 @@ namespace fb.fbdoc
 
 		res = _closeList( ctx )
 		
-		cellData = paramsTb->GetParam( "cells" )
+		cellData = UnescapeHtml( paramsTb->GetParam( "cells" ) )
 		strCols = paramsTb->GetParam( "columns" )
 
 		n = _index_cells( cellData, @cellTb )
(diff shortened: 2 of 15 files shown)


swurzinger:
rebased to HEAD

dkl:
Looks good to me in general (I'm not very familiar with fbdoc code base though), nice work.

I tried to see what differences this causes, so I compared HTML files produced by old and new versions. There seems to be lots more proper escaping, which is great.

On KeyPgPrintusing (maybe others too) the table formatting needs to be fixed: currently there are cells containing `&amp`, and with the `;` cell separator it becomes `&amp;text`, which is unescaped to a single `&text` cell instead of two cells.

It may be useful to search the *.wakka files in doc/manual/cache for the HTML escapes to find such cases... `&amp` only appears on KeyPgPrintusing (and TestPage) currently though, so it seems like a small issue.

swurzinger:
I've searched the wakka files for further invalid escape sequences and found only a few & fixed them.

While testing I found another encoding issue, this time utf-8 related: the latest commit should fix that by adding the "charset" meta property to the html document. An example page that was broken in the html/chm previously and should work now is https://freebasic.net/wiki/wikka.php?wakka=CptAscii

jayrm:
Looks good to me.  Hope you don't mind, I pushed a minor change to this PR that is related, though it has to do with links.  There is a tool called fbchkdoc/rebuild that parses and rewrites wakka files.  The theory is that if all the information is correctly captured from wakka script, then should be possible to rewrite (emit) as wakka with only whitespace and/or newline changes.  Can try it on a copy of the files by copying everything from ./manual/cache to ./manual/cache.web,  run `../fbchkdoc/rebuild -web+ @PageIndex.txt` and compare the files between cache and cache.web.

PR #50, opened 2017-10-25
Change negation UOP to always produce signed result

dkl:
The unary generation now always gives a signed result. Previously it only produced a signed result in case the operand was a constant, so `-1u` gave an Integer, but `-myuintvar` gave a UInteger. (#841)

It's different from C's rules, but as far as I know it was intentionally done this way for FB. And I think it's important for constant and non-constant operations to give the same result. Obviously this change breaks backwards compatibility.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -46,6 +46,7 @@ Version 1.06.0
 - #847: Windows API binding: Fixed winnt.bi SECURITY_*_AUTHORITY initializers
 - #851: '#LINE line filename' only changed the source filename in error messages, not in debug info
 - #832: Fix bug allowing QB style suffixes on all keywords, regardless of -lang
+- #841: The unary negation operator gave different result signedness when used on constant instead of non-constant expression. Now the result is always signed.
 
 
 Version 1.05.0
--- a/src/compiler/ast-node-uop.bas
+++ b/src/compiler/ast-node-uop.bas
@@ -224,9 +224,28 @@ function astNewUOP _
 			subtype = o->subtype
 		end if
 
-	'' with SGN(int) the result is always a signed integer
+	'' with SGN(int) and [u]integer negation the result is always a signed integer
 	case AST_OP_SGN
 		if( typeGetClass( o->dtype ) = FB_DATACLASS_INTEGER ) then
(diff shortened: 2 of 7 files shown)


jayrm:
I don't think it breaks backwards compatibility in a meaningful way.  I tried writing a test to see differences before and after.  I am having a tough time imagining what kind of program is going to become broken because of the change.  Datatype changes but internal bits are same.

```
'' Unary neg op test

function ineg( byval i as uinteger ) as integer
	function = -i
end function

function uneg( byval i as uinteger ) as uinteger
	function = -i
end function

#macro test( x )
	print x, -(x), -(-(x))
	print x, ineg(x), ineg(ineg(x))
	print x, uneg(x), uneg(uneg(x))
	print
#endmacro

dim a as uinteger = 1
dim b as uinteger = -1
dim c as uinteger = ineg(1)
dim d as uinteger = uneg(1)

test( 1 )
test( -1 )
test( a )
test( b )
test( c )
test( d )

/'

BEFORE

 1            -1             1
 1            -1             1
 1            4294967295    1

-1             1            -1
-1             1            -1
-1            1             4294967295

1             4294967295    1
1             -1             1
1             4294967295    1

4294967295    1             4294967295
4294967295     1            -1
4294967295    1             4294967295

4294967295    1             4294967295
4294967295     1            -1
4294967295    1             4294967295

4294967295    1             4294967295
4294967295     1            -1
4294967295    1             4294967295


AFTER (unary neg returns signed type)

 1            -1             1
 1            -1             1
 1            4294967295    1

-1             1            -1
-1             1            -1
-1            1             4294967295

1             -1             1
1             -1             1
1             4294967295    1

4294967295     1            -1
4294967295     1            -1
4294967295    1             4294967295

4294967295     1            -1
4294967295     1            -1
4294967295    1             4294967295

4294967295     1            -1
4294967295     1            -1
4294967295    1             4294967295

'/
```

dkl:
Ok, thanks for reviewing. This is one of those changes that I wouldn't just want to push without asking for feedback, especially because it breaks backwards compatibility (even if only a bit), and also because I'm no longer as active as before.

PR #51, opened 2017-10-25
travis: Do fbc debug build for better testing

dkl:
Enable assertions and -exx during travis build, for better testing.


jayrm:
I typically build everything with `-g -exx`.  That is my preferred build, so I think is good idea.

I'm not familiar with travis.   What is main purpose 'continuous-integration/travis-ci/pr' ?  Development only?  For a release might be worthwhile to do both debug and non-degug build and test.  Occasionally there is a bug that appears in one but not the other.  If can only choose one, because of resources, I like `-g -exx` change.

dkl:
Yea, the travis CI build is for testing purposes during development. I like the idea of building both release and debug versions; will give it a try.

PR #56, opened 2017-11-26
fbdoc & tools update Nov 2017

jayrm:
fbdoc, & tools plus wiki snapshot update for Nov 2017 includes the following:
- wiki snapshots up to Nov 26, 2017
- fbdoc: doc/manual/cache/*.wakka files always checkout as LF (need to delete and revert this dir to see change on Win PC's).
- updated templates/default/keywords.lst
- removed hard coded path from fbdoc tools (was ../manual) and is now an option in the INI.  Build will probably break until you update the .ini file with "manual_dir=../manual"
- extracted all the samples from wiki and updated examples/manual directories

Many files get touched, but overall, changes are minor.


doc/fbchkdoc/mkprntoc.bas
```diff
@@ -296,6 +330,10 @@ while( command(i) > "" )
 			cache_dir = web_cache_dir
 		case "-dev+"
 			cache_dir = dev_cache_dir
+		case "-with-tut-pages"
+			opt_with_tut_pages = true
+		case "-with_dev-pages"
```
dkl: the `_` in `-with_dev-pages` looks like a typo here

doc/libfbdoc/fbdoc_keywords.bas
```diff
@@ -79,7 +79,13 @@ namespace fb.fbdoc
 
 		keyworddata = LoadFileAsString( sFileName )
 
-		keywordhash.alloc( 1000 )
+		keywordhash.alloc( 2000 )
+
+		if( keyworddata = "" ) then
+		exit function
```
dkl: This `exit function` is probably not supposed to be there

doc/manual/templates/default/keywords.lst
```diff
@@ -209,6 +220,7 @@ ImageConvertRow
 ImageDestroy
 ImageInfo
 Imp
+Implememts
```
dkl: Typo

doc/manual/templates/default/keywords.lst
```diff
@@ -20,11 +21,13 @@ $Dynamic
 $Include
 $Lang
 $Static
+Abstract
 Assert
 AssertWarn
 Abs
 Access
 Acos
+Add
```
dkl: `add` isn't an FB keyword, or is it?

doc/fbchkdoc/fmtcode.bas
```diff
@@ -135,17 +136,31 @@ private function _emitCode ( byref text as string ) as string
...
+''
+function FormatFbCodeLoadKeywords( byref filename as string ) as boolean
+
 	static kw_loaded as integer = FALSE
 
 	if( kw_loaded = FALSE ) then
 		'' Load the keywords, this will allow keywords to be cased.
-		fbdoc_loadkeywords( "../manual/templates/default/keywords.lst" )
+
+		if( filename > "" ) then
```
dkl: How about rewriting that as:
```
if( filename = "" ) then
    filename = default_ManualDir + "templates/default/keywords.lst"
end if
function = ( fbdoc_loadkeywords( filename ) <> 0 )
```

jayrm: I like your suggestion, it flows better.  It also highlights that I wrote most of this code before const qualifiers and will have to revisit the code base for an overall update.

doc/fbchkdoc/mkprntoc.bas
jayrm: if I type it wrong consistently, is it still a typo?, lol.  OK, fixed.

doc/libfbdoc/fbdoc_keywords.bas
jayrm: Correct.  I did get the printed message at one point during testing... weird.  Thanks.

doc/manual/templates/default/keywords.lst
jayrm: Add is a graphics keyword documented at https://www.freebasic.net/wiki/wikka.php?wakka=KeyPgAddGfx

doc/manual/templates/default/keywords.lst
jayrm: Fixed.  Thanks.

doc/manual/templates/default/keywords.lst
dkl: Oh, oops, missed it.

doc/fbchkdoc/fmtcode.bas
dkl: Hmm, yea, modifying the Byref argument may be unexpected, so a copy is needed. However, as far as I saw, FormatFbCodeLoadKeywords() is never called with an empty string at the moment, so it may not even need to handle the case.

doc/fbchkdoc/fmtcode.bas
```diff
@@ -145,10 +145,9 @@ function FormatFbCodeLoadKeywords( byref filename as string ) as boolean
 		'' Load the keywords, this will allow keywords to be cased.
 
 		if( filename > "" ) then
```
dkl: Currently the logic seems broken: `if filename > ""` should be `if filename = ""`

jayrm: Yes, you are correct.  I'm so sorry.  That was very careless of me.  My test method was poor (false success), and I didn't test all cases.  

Personal observation about my own code, having not looked at it in years: these doc utility programs have a lot of duplicated start-up code, and the code is too complex for what it accomplishes.  That is an area for improvement for next time.  Otherwise, I am finding the utilities are still relevant for maintaining the wiki, even though they may not be most user friendly.

dkl: Yea, fbdoc & co have been crucial and useful for me too for working on FB in the past years. I had wanted to make some improvements to them here and there, but never got around to completing any. It's nice that they worked generally without problems ever since I started.

Issue #60, opened 2018-01-20
Can't link to zlib 1.2.11


novavoon:
C:\FB\fbc -s console "zlib.bas"
C:\FB\bin\win32\ld.exe: skipping incompatible ./zlib1.dll when searching for -lzlib1
C:\FB\bin\win32\ld.exe: skipping incompatible ./zlib1.dll when searching for -lzlib1
C:\FB\bin\win32\ld.exe: cannot find -lzlib1

jayrm:
What version of fbc?  Do you have a 64-bit zlib installed where linker is expecting 32-bit?

novavoon:
nevermind I forgot that I am using MinGW64 to compile my zlib so I expect it to be a 32bit library. Problem solved after using 64bit version of fbc to compile.

PR #61, opened 2018-01-20
Updated zlib.bi to use the 3 common library names

takase1121:
now zlib.bi can use either
* zlib1
* z
* libzlib 

to include the library. This fixes issues when trying to include zlib in projects.

to use a specific library name:
* for "zlib1"   : `#define ZLIB_USE_ZLIB1`
* for "z"         : `#define ZLIB_USE_Z`
* for "libzlib" : `#define ZLIB_USE_LIBZLIB`

If none of them are defined, the default choice is "z"
The `#define` has to be done before `#include "zlib.bi"`

### DIFF
--- a/inc/zlib.bi
+++ b/inc/zlib.bi
@@ -35,7 +35,22 @@
 
 #pragma once
 
+#ifdef ZLIB_USE_ZLIB1
+#define NAMEDEF 
+#inclib "zlib1"
+#endif
+#ifdef ZLIB_USE_Z
+#define NAMEDEF
 #inclib "z"
+#endif
+#ifdef ZLIB_USE_LIBZLIB
+#define NAMEDEF
+#inclib "libzlib"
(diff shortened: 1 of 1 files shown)


jayrm:
Assuming this is for win32 (based on issue #60 that was submitted), this is not the correct solution.  Should not be mixing different versions of zlib.bi header, zlib.dll or zlib1.dll

What version of fbc?  What operating system?

need to have zlib version 1.2.8 import library
./lib/win32/libz.dll.a
and zlib version 1.2.8 dll in current directory or path.
zlib1.dll

Best solution would be to update zlib.bi header for zlib version 1.2.11 dll

For more information about zlib, name change, and incompatible libraries, see https://www.zlib.net/DLL_FAQ.txt

novavoon:
I see. Thanks anyways!

PR #62, opened 2018-01-22
Fix C++ mangling for BYREF parameters with built-in types

dkl:
hAbbrevFind() skipped the lookup for reference-to-built-in types. (and potentially even more)

However, references are considered for compression too by the Itanium C++ ABI. We already do hAbbrevAdd() for such cases too.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -47,6 +47,7 @@ Version 1.06.0
 - #851: '#LINE line filename' only changed the source filename in error messages, not in debug info
 - #832: Fix bug allowing QB style suffixes on all keywords, regardless of -lang
 - #841: The unary negation operator gave different result signedness when used on constant instead of non-constant expression. Now the result is always signed.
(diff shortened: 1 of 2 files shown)


jayrm:
I have a unit test to add that will make some basic checks for c++ name mangling.  I can't push to this pull request, I'm going to try adding a pull request that builds on this one.

jayrm:
I think this PR can be merged in. Then, the test for the c++ name mangling is completed, for now, in PR #63

The only test that couldn't pass on all platforms was mapping some fbc integer type to g++'s long int type. I saw the notes in symb-mangling.bas:hMangleBuiltInType().

If we are mangling names for c++ specifically, and did want cross-platform consistency, then the overloading rules for built-in types would need to follow g++ for the c++ mangling only. Otherwise, if we are compiling and linking fbc only, then we don't really care what the mangling is, so long it is unique.

dkl:
Thanks, good idea to add the test.

PR #63, opened 2018-01-28
log test for PR#62 -- Fix C++ mangling for BYREF parameters with built-in types

jayrm:
intent is to add basic testing for c++ name mangling

### DIFF
--- a/tests/cpp/fbc-mangle.bas
+++ b/tests/cpp/fbc-mangle.bas
@@ -9,8 +9,12 @@
 	type cxxlongint as integer
 
 	/' !!!
-		cxxlongint will never work on win64, fbc custom mangles INTEGER and there is 
+		cxxlongint will currently fail on win64, fbc custom mangles INTEGER and there is 
 		no other data type that will return the "m" and "l" suffixes for c's long int.
+		The specific test that fails is skipped over using and and #ifdef.  To have a basic
+		datatype on win64 that can map to g++ long int, one possible solution would be
+		change the overloading allowed for the c++ mangled names onlym what would have to
+		be added to fbc.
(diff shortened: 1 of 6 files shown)


jayrm:
I think this PR can be merged in.  Then, the test for the c++ name mangling is completed, for now, in PR #62

The only test that couldn't pass on all platforms was mapping some fbc integer type to g++'s long int type.  I saw the notes in symb-mangling.bas:hMangleBuiltInType(). 

If we are mangling names for c++ specifically, and did want cross-platform consistency, then the overloading rules for built-in types would need to follow g++ for the c++ mangling only.  Otherwise, if we are compiling and linking fbc only, then we don't really care what the mangling is, so long it is unique.

tests/cpp/fbc-mangle.bas
```diff
@@ -0,0 +1,215 @@
...
+'' test mapping of basic data types between fbc and g++
+'' with c++ name mangling
+
+#ifdef __FB_64BIT__
+
+	type cxxint as long
+	type cxxlongint as integer
+
+	/' !!!
+		cxxlongint will never work on win64, fbc custom mangles INTEGER and there is
```
dkl: Does that mean that the test will fail to compile on win64? If so we should probably disable it with #ifdefs etc. for that case.

dkl: Ah ok, that must be why that part is commented out below. I suppose it could use `#if (not defined(__FB_64BIT__)) or defined(__FB_UNIX__)` or similar.

jayrm:
That's a good idea, then it still test the assertions on systems where it should work.

PR #64, opened 2018-01-29
Update contrib/release/build.sh for fbc 1.06

jayrm:
This updates and improves on the contrib/release/build.sh script for creating releases:
- added some command line argument parsing
- the --offline option (WIP) is intended to only use files that are locally available without connecting to internet
- the --repo url --remote name options allow building a release from an alternate repo other than freebasic/fbc
- copy the directx headers from the build system
- updated the packages needed for djgpp
- bootstrap from fbc 1.05.0

Tested:
- on win7 to build fbc win 32-bit packages
- on win7 to build fbc win 64-bit packages
- on slackware 14.1 to build fbc linux 64-bit packages

The dos version only partially tested.  I don't have a build environment that will work with.  The updates for fbc dos version are pasted in from commands manually typed in when testing on winxp & win98 VM systems.

No guarantees that this produces a valid distribution, however, the script does complete.

### DIFF
--- a/contrib/release/build.sh
+++ b/contrib/release/build.sh
@@ -219,8 +219,8 @@ win32-mingworg)
 	download_extract_mingw mpfr-3.1.2-2-mingw32-dll.tar.lzma
 
 	# Add ddraw.h and dinput.h for FB's gfxlib2
-	download dx80_mgw.zip http://alleg.sourceforge.net/files/dx80_mgw.zip
-	unzip ../input/dx80_mgw.zip include/ddraw.h include/dinput.h
+#	download dx80_mgw.zip http://alleg.sourceforge.net/files/dx80_mgw.zip
+#	unzip ../input/dx80_mgw.zip include/ddraw.h include/dinput.h
 
 	# Work around http://sourceforge.net/p/mingw/bugs/2039/
 	patch -p0 < ../mingworg-fix-wcharh.patch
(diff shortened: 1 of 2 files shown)


contrib/release/build.sh
```diff
@@ -257,7 +353,7 @@ linux*)
 
 	# fbc sources
 	cp -R ../input/fbc fbc
-	cd fbc && git reset --hard "$fbccommit" && cd ..
+	cd fbc && git reset --hard "$remote_name"/"$fbccommit" && cd ..
```
dkl: If `$fbccommit` is a SHA1 hash then it will break, I think something like `origin/7fbad71f6508342a57acbf7527f456c794bae2b0` doesn't work.

jayrm:
Ok, I changed "$remote_name"/"$fbccommit" back to "$fbccommit" so should can use the sha-1 commit id again. yea, I was only thinking of branches and tags.

Example of use: if I have repo on a file share, and it has a branch called 'test-pkg' I can build from that:
`$ ./build.sh win32 other/test-pkg --url file:////server/repo/fb.git`
`$ ./build.sh win32 myrepo/test-pkg --url file:////server/repo/fb.git --remote myrepo`

and also with some commit id:
`$ ./build.sh win32 9abc123 --url file:////server/repo/fb.git`

That works for me.  It's handy if trying to test or experiment with the build packages before pushing to public, and the build itself depends on files in the repo, like scripts or makefiles.

PR #69, opened 2018-03-10
Macros invoked without parentheses and new builtin macros

av1ctor:
Added the builtin macros: __FB_UNIQUEID_PUSH__(), __FB_UNIQUEID__(), __FB_UNIQUEID_POP__(), __FB_ARG_LEFTOF__(), __FB_ARG_RIGHTOF__(), __FB_JOIN__(). They make it easier to create complex macros and syntax sugars.

Added option to call macros without using parentheses, allowing a more BASICish syntax with some macros.

Check the examples/misc/trycatch and examples/misc/generics to see how they could be used.

### DIFF
--- a/examples/misc/generics/list.bi
+++ b/examples/misc/generics/list.bi
@@ -1,0 +1,66 @@
+#macro list(T)
+	__fb_join__(list_,__fb_arg_rightof__(T,OF))
+#endmacro
+
+#macro list_of(T)
+	list_##T
+#endmacro
+
+#macro imp_list_of(T)
+#if not defined(list_##T)	
+	type list_item_##T
+		value as T
+		next_ as list_item_##T ptr
+	end type
+
+	type list_##T
+		declare constructor(initial as integer = 0)
+		declare destructor()
+		declare sub add(v as T)
+		declare operator[](idx as integer) as T
+	private:
+		head as list_item_##T ptr
+		tail as list_item_##T ptr
+	end type
+
+	constructor list_##T(initial as integer)
+		this.head = 0
+		this.tail = 0
+	end constructor
+	
+	destructor list_##T()
+		do while head <> 0
+			var next_ = head->next_
+			deallocate(head)
+			head = next_
+		loop
+	end destructor
+	
+	sub list_##T.add(v as T)
+		dim as list_item_##T ptr node = allocate(len(list_item_##T))
+		node->value = v
+		if tail <> 0 then
+			tail->next_ = node
+		else
+			head = node
+		end if
+		node->next_ = 0
+		tail = node
+	end sub
+	
+	operator list_##T.[](idx as integer) as T
+		var i = 0
+		var node = head
+		do while node <> 0
+			if i = idx then
+				return node->value
+			end if
+			node = node->next_
+			i += 1
+		loop
+		'return default(T)
+	end operator
+#endif
+#endmacro
+
+
--- a/src/compiler/pp-define.bas
+++ b/src/compiler/pp-define.bas
@@ -17,17 +17,6 @@
 				   LEXCHECK_NOQUOTES or _
 				   LEXCHECK_NOSYMBOL)
 
-type LEXPP_ARG
-	union
-		text		as DZSTRING
-		textw		as DWSTRING
-	end union
-end type
-
-type LEXPP_ARGTB
-	tb(0 to FB_MAXDEFINEARGS-1) as LEXPP_ARG
-end type
-
 ''::::
 sub ppDefineInit( )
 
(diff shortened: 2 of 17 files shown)


jayrm:
v1c, looks really cool.  Examples work for me on win32 & dos (-gen gas).  I am confident test suite will pass with fbcunit.bi::TEST() macro after I get the tests rewritten.

For me, I had errors compiling examples/misc/trycatch/ on both slackware-64bit & win7-64bit (-gen gcc):
```
test.bas(21): error 148: Suffixes are only valid in -lang fb or deprecated or fblite in 'try'
test.bas(21): error 148: Suffixes are only valid in -lang fb or deprecated or fblite in 'try'
test.bas(21): warning 3(1): Passing different pointer types, at parameter 1 of SETJMP()
... and so on
```
Not sure, but looks like -gen gcc uses '$' in identifiers.  Possibly related to fix for No. 832 on sf.net?

Example:
```
'' a.bas
__FB_UNIQUEID_PUSH__(a)
#print __FB_UNIQUEID__(a)

/'
OUTPUT:

D:\fb.git>fbc-32 a.bas
Lt_0004

D:\fb.git>fbc-64 a.bas
a.bas(3): error 148: Suffixes are only valid in -lang fb or deprecated or fblite in '#print __FB_UNIQUEID__(a)'
'/
```

av1ctor:
Hi, Jeff. 

Fixed. The problemas was that in -gen gcc mode symbUniqueId() added a $ prefix in the symbol name returned.

What I really wanted to add was a regex-like macro params, like:

#macro qb_line(/({expr}, {expr})-({expr}, {expr}), {expr}, [b|bf]/) 
 #if __fb_arg__(5) = ""
    gfx.line(__fb_arg__(0), __fb_arg__(1), __fb_arg__(2), __fb_arg__(3), __fb_arg__(4))
#elseif __fb_arg__(5) = "B"
    gfx.box(__fb_arg__(0), __fb_arg__(1), __fb_arg__(2), __fb_arg__(3), __fb_arg__(4))
#elseif __fb_arg__(5) = "BF"
    gfx.boxf(__fb_arg__(0), __fb_arg__(1), __fb_arg__(2), __fb_arg__(3), __fb_arg__(4))
#else
   #error "Syntax Error: Expected B or BF at parameter 6 of qb_line"
#endif
#endmacro

So you could do:

var x = 0, y = 0, c = 15
qb_line (x+10, y+10)-(x+100, y+100), c, BF

Maybe next time :)

examples/misc/generics/list.bi
```diff
@@ -0,0 +1,66 @@
...
+#endmacro
+
+#macro imp_list_of(T)
+#if not defined(list_##T)	
+	type list_item_##T
+		value as T
+		next_ as list_item_##T ptr
+	end type
+
+	type list_##T
```
dkl: It would probably be good to also define the copy constructor and `operator let` (or at least declare them and leave them unimplemented).

examples/misc/generics/list.bi
```diff
@@ -0,0 +1,66 @@
...
+		declare constructor(initial as integer = 0)
+		declare destructor()
+		declare sub add(v as T)
+		declare operator[](idx as integer) as T
```
dkl: This might be a candidate for `byref` result (`declare operator[](idx as integer) byref as T`)

src/compiler/pp-define.bas
```diff
@@ -68,25 +57,34 @@ private function hLoadMacro _
...
-    dim as FBTOKEN t = any
+	dim as FBTOKEN t = any
     dim as LEXPP_ARGTB ptr argtb = any
 	dim as integer prntcnt = any, num = any, reached_vararg = any, is_variadic = any
     dim as zstring ptr argtext = any
     static as string text
 
 	function = -1
 
+	var isParentLess = false
```
dkl: maybe `hasParens` to avoid the double negative?

src/compiler/symb-define.bas
```diff
@@ -124,11 +122,11 @@ private function hDefErr_cb ( ) as string
 	function = str( res )
 end function
 
-private function hDefLang_cb ( ) as string
+private function hDefLang_cb() as string static
```
dkl: Is `static` really useful here? As far as I remember, it only affects normal variables, not temps or function results.

src/compiler/symb-define.bas
```diff
@@ -161,17 +159,200 @@ private function hDefFpmode_cb ( ) as string
...
+		*errnum = FB_ERRMSG_ARGCNTMISMATCH
+		return ""
+	end if
+
+	dim tokens() as string
+	var numtoks = hStr2Tok(arg, tokens())
+	
+	if( numtoks = 0 ) then
+		*errnum = FB_ERRMSG_SYNTAXERROR
+		return ""
```
dkl: It looks like this error path would leak `arg` and `sep`.

src/compiler/symb-define.bas
```diff
@@ -161,17 +159,200 @@ private function hDefFpmode_cb ( ) as string
...
```
dkl: same here

jayrm:
Imagine this feature to invoke macro expansion without parentheses, plus hook the macro expansion to a user plugin dll procedure loaded by fbc at run-time.... If the API was not too complex, any user could extend fbc features. Well, some users anyway.

PR #72, opened 2018-04-15
fbc: fix bug #844 duplicated type (struct) in -gen gcc

jayrm:
See: #844 gen gcc: C code conflict between array-descriptor/UDT structures when declare
https://sourceforge.net/p/fbc/bugs/844/


jayrm:
I'm going to try stacking a few pull requests.  Not sure if it is worthwhile or not, so if it doesn't work out, I won't do it again.  Each pull request is based on the previous pull request.  From what I have read, if all goes well,  the last pull request is that first to get merged.

PR #74, opened 2018-04-16
Fix #872 broken boolean bitfield runtime assignments from unsigned values

jayrm:
#872: Fix broken boolean bitfield runtime assignments from unsigned values
See https://sourceforge.net/p/fbc/bugs/872/

### DIFF
--- a/src/compiler/ast-node-misc.bas
+++ b/src/compiler/ast-node-misc.bas
@@ -431,7 +431,7 @@ private function astSetBitfield _
 
 	'' boolean bitfield? - do a bool conversion before the bitfield store
 	if( symbGetType( bitfield ) = FB_DATATYPE_BOOLEAN ) then
-		if( r->class <> AST_NODECLASS_CONV ) then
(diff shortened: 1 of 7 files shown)


src/compiler/ast-node-misc.bas
```diff
@@ -433,6 +433,10 @@ private function astSetBitfield _
 	if( symbGetType( bitfield ) = FB_DATATYPE_BOOLEAN ) then
 		if( r->class <> AST_NODECLASS_CONV ) then
```
dkl: How about:
```
if( (r->class <> AST_NODECLASS_CONV) or (astGetFullType( r ) <> FB_DATATYPE_BOOLEAN) ) then
	r = astNewCONV( FB_DATATYPE_BOOLEAN, NULL, r )
end if
```

jayrm: Done

jayrm:
The PR stacking seemed to work fine; after merging this PR, the other two automatically merged and closed.  My opinion: it would be better if the latest PR was diff'ed against dependant PR.  I found it confusing that the latest PR showed all the commits and changes for the previous PRs.

PR #76, opened 2018-04-20
Allow overload operator SQR() 

jayrm:
- allows SQR() operator to be overloaded like ABS, SGN, etc.
Additionally,
- alter floating point tests under djgpp with larger tolerance (epsilon) in checks
- alter floating point tests when compiling with -fpu sse -fpmode fast with larger tolerance (epsilon) in checks


jayrm:
The only change I am not sure about is the check to optimize in ast-optimize.bas.  Previously, there was no check for the type of the SQR() result on the right hand side.  The parser would have checked types before it got to here.  My other choice was only allow single data type.

dkl:
The point was to prevent the UDT case from being optimized, right? It looks good to me. Possibly at that point the overloading already happened and the AST would have a CALL node for the UDT case, not an UOP, but it seems like even then checking the rhs type wouldn't hurt.

PR #77, opened 2018-04-22
Fix #875 for -gen gas -fpu sse

jayrm:
Fix #875 for -gen gas -fpu sse where boolean variable to single/double conversion gives wrong sign.


jayrm:
When testing fbc with different math options, found this bug related to #875, but for a different reason.  SSE emitter can't handle any results on FPU stack expect for a procedure return.   Maybe not the best code gen, but was simplest to implement without messing with the IR.  Uses the x86 boolean to float emitter proc int, then transfers to sse register.

PR #78, opened 2018-04-22
fbc-tests: fix inline-asm test to work on all platforms

jayrm:
This makes minor change to tests/quirk/inline-asm.bas to test on all platforms:
test with:
- dos, win32 -gen gas 
- dos, win32, -gen gcc
- win64, linux64, -gen gcc

### DIFF
--- a/tests/quirk/inline-asm.bas
+++ b/tests/quirk/inline-asm.bas
@@ -9,17 +9,27 @@ sub quirk_inline_asm_f1 cdecl alias "quirk_inline_asm_f1"()
 end sub
 
 private sub test_proc
+
+	#define quirk_name quirk_inline_asm_f1
+
+	#ifndef __FB_64BIT__
+		#if __FB_GCC__
+			#undef quirk_name
+			#define quirk_name _quirk_inline_asm_f1
+		#endif
+	#endif
+
 	CU_ASSERT( f1calls = 0 )
-	asm call quirk_inline_asm_f1
+	asm call quirk_name
 	CU_ASSERT( f1calls = 1 )
 
 	dim f1address as any ptr
 	asm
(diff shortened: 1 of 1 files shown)


jayrm:
I am unable to test on linux 32 bit currently.  I am also hoping the results of this test might give clues as to why 32bit -gen gcc targets have name mangling (leading underscore) linking problems.

dkl:
The underscore is only needed by the ABI for dos/win32, but for the C backend fbc generally doesn't need to add it, because gcc already does it - except in inline asm. I guess that's where fbc gets it wrong.

https://github.com/freebasic/fbc/blob/791f00de655674c42276acbdda4bff3d8e27ebf8/src/compiler/fb.bas#L381

https://github.com/freebasic/fbc/blob/791f00de655674c42276acbdda4bff3d8e27ebf8/src/compiler/symb-mangling.bas#L576

https://github.com/freebasic/fbc/blob/791f00de655674c42276acbdda4bff3d8e27ebf8/src/compiler/ir-hlc.bas#L456

I'm guessing the bug is here somewhere: https://github.com/freebasic/fbc/blob/791f00de655674c42276acbdda4bff3d8e27ebf8/src/compiler/ir-hlc.bas#L3301
For the -asm intel case, this should probably use `hGetMangledNameForASM( n->sym, TRUE )` instead of the raw `symbGetMangledName( n->sym )`.

Simpler test case:
```
extern "C"
    sub foo()
        print "foo"
    end sub
end extern

asm call foo
```

`fbc -target i686-w64-mingw32 -gen gcc -rr` generates `call foo` for this (wrong), while -gen gas generates `call _foo` (correct).

jayrm:
Closed in favour of pull request #79

PR #79, opened 2018-04-23
Fix proc name mangling in inline asm under -gen gcc -asm intel

dkl:
This should fix tests/quirk/inline-asm.bas to build for win32. @jayrm, could you give it a try?


jayrm:
Yes, tested -gen gcc for win32 & dos.  Looks good.

PR #80, opened 2018-04-28
Fix #878: select case as const

jayrm:
fix #878: fbc-64bit hangs on 'Case True' + 'Case False' inside 'Select Case As const'

This change was started based on sf bug #878, and includes several additional changes:
- jump tables generated by SELECT CASE AS CONST use bias/span internally, instead of minval/maxval
- SELECT CASE AS CONST respects data type used when parsing CASE range values
- CASE range values can be mix of signed & unsigned constants
- additional tests added in the test-suite

Initial discussion at https://www.freebasic.net/forum/viewtopic.php?f=3&t=26621


jayrm:
I was unable to test LLVM backend (although I did try).  In fbc (and C) I am certain how integer math wraps around for signed/unsigned value in a register; not so sure both LLVM.  I suspect that LLVM emitter will work for most CASE ranges, but may fail if the CASE value ranges are signed and close to +/-2^31, or unsigned and close to 2^32.

src/compiler/parser-compound-select-const.bas
```diff
@@ -209,49 +219,50 @@ sub cSelConstStmtNext( byval stk as FB_CMPSTMTSTK ptr )
...
-		dim as ulongint value = cConstIntExpr( cExpression( ), symbGetType( stk->select.sym ) )
-
-		var minval = stk->select.const_.minval
-		var maxval = stk->select.const_.maxval
+		dim as ulongint value = cConstIntExpr( cExpression( ), stk->select.const_.dtype ) - stk->select.const_.bias
 
+		'' first case?
+		if( swtbase = ctx.base ) then
+			stk->select.const_.bias = value - FB_MAXJUMPTBSLOTS
+			value -= stk->select.const_.bias
```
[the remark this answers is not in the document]
jayrm: OK.  With your suggestion, I think I made it more obvious what's happening...

src/compiler/parser-compound-select-const.bas
```diff
@@ -298,13 +307,25 @@ sub cSelConstStmtEnd( byval stk as FB_CMPSTMTSTK ptr )
     '' emit comp label
     astAdd( astNewLABEL( stk->select.cmplabel ) )
 
+	'' re-bias case values and set max value
+	dim as ulongint span = 0
```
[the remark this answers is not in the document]
jayrm: Span is simply the difference between the upper value and lower value of the range; the algebraic length of the interval.  A synonym for span might be interval length, or distance.  My knowledge of the term "span" comes from instrumentation.  For example: a thermometer with a range from -40 deg C to 120 deg C, has a zero offset (bias) of -40 deg C and a span of 160 deg C.

With SELECT CASE AS CONST we are testing a value against a closed interval [minval, maxval].  By converting to [bias, bias+span], the comparison can be rewritten as (value-bias) tested against [0, span].  Or simply as (value-bias)<=span.  Formula works for real numbers also (but obviously not the "wrap" around integer property).

Only when the range is finally expressed as a jump table do we count the integers contained in the closed range.  For example, in our use, a span of zero (0), minval=maxval, contains exactly 1 countable integer.  If there is no range, then span is undefined.

jayrm:
I believe this patch fixes the bug reported; the compiler won't crash or hang.  The current discussion on the forum is that it also changes expected behaviour of SELECT CASE AS CONST compared to SELECT CASE.  To fully address the issue, additional code is needed to check/warn on ranges, and must not CAST range values to the SELECT CASE AS CONST data type.

countingpine:
There is a question I've mentioned on the forum, essentially about how case values should be handled when they're out of range of the datatype.

Unless it's a controversial decision, I think I'd like to implement an explicit range-check warning, rather than cast the case values down to the source type.

The way I'm thinking of it now, there are two basic kinds of situation with out-of-bounds values:
- Assignment: `dim b as byte = &hff` - this requires a "truncation", but gives the expected outcome
- Comparison: `if b = &hff then ...` - this will always be false, and reveals that the value is not "signedness-agnostic".

I think Case values (and equality checks in general) are of the latter kind, and shouldn't do any special handling for small signed variables.

I've not put any code up yet around this issue, but I can take a look at it.  Although I don't want to step on anyone else's work..

jayrm:
I don't think it is controversial, just different, and sets a precedent for what users might be expect in other areas of compilation. 

I added cConstIntExprRanged() for SELECT CASE AS CONST.  It's called and used the same way as cConstIntExpr() except it produces an overflow warning if the (un)signed constant does not fit in the (un)signed type.

Ugh, last build failed, due to warnings changed.

countingpine:
Does cConstIntExprRanged() handle 64-bit [u]integers properly? I see code that specifically checks for the [U]LONGINT datatype.

jayrm:
It handles them properly.  Converting between longint and ulongint is a special case.  The compiler uses longint to store both signed/unsigned integers.  That's fine for all smaller sized integers.  But for longint/ulongint need like 64.5 bits for the comparison.  So the only way is to check the [u]longint datatype.  Another way would have been to test if the highest bit was set.  Would still be a special case.

jayrm:
In this last update I added a fairly extensive test for the warnings.  It's added under SELECT CASE AS CONST, but only because that is the only place where cConstIntExprRanged() is used.

The only irregularity I found is the warning generated by astBuildJMPTB().  It occurs when:
1) the SELECT CASE AS CONST datatype is 32-bit or smaller, and
2) we are building for a 32-bit target, and
3) one of the CASE ranges exceeds the 32-bit data type (therefore bias exceeds 32bits)

In such a scenario, the compiler gives 2 warnings, first when the CASE [u]longint is encountered, and then at END SELECT when the jump table calculation subtracts the over 32-bit bias.

I considered hiding the warnings when astBuildJMPTB() is called, but not sure if there are other warnings could be be generated.

jayrm:
I think we can merge this in.  I am satisfied that it has been tested.  I tested llc & llvm 6.0 32-bit, and this patch appears to work.  There are other llvm compiling issues, but specifically, the `select case as const` appears correct enough.

A case I was specifically concerned about is where the bias must be 64 bit due the case values, but the data type is 32bit:
```
dim i as ulong = 4294967296
select case as const i
case 4294967296
	end 0
case else
	end 1
end select
end 2
```
the LL asembler output we get:
```
	; jmptb LT_0006
	%vr7 = load i32, i32* %LT_0006.4
	switch i32 %vr7, label %.Lt_0008 [
		i32 4294967296, label %.Lt_0007
	]
```
Even though we get compiler overflow warnings, and the  constant emitted is greater than 32 bit, the assembler appears to read the constant value and truncate, so the bit pattern matches as expected.

src/compiler/parser-decl-symbtype.bas
```diff
@@ -29,6 +29,85 @@ function cConstIntExpr _
...
+	static range( FB_SIZETYPE_BOOLEAN to FB_SIZETYPE_UINT64 ) as RANGEVALUES = _
+		{ _
+			/' FB_SIZETYPE_BOOLEAN '/ ( -1, 0, 0 ),  _
+			/' FB_SIZETYPE_INT8    '/ ( -128, 127, 127 ), _
+			/' FB_SIZETYPE_UINT8   '/ ( 0, 127, 255 ), _
+			/' FB_SIZETYPE_INT16   '/ ( -32768, 32767, 32767 ), _
+			/' FB_SIZETYPE_UINT16  '/ ( 0, 32767, 65535 ), _
+			/' FB_SIZETYPE_INT32   '/ ( -2147483648ll, 2147483647ll, 2147483647ull ), _
+			/' FB_SIZETYPE_UINT32  '/ ( 0, 2147483647ll, 4294967295ull ),  _
+			/' FB_SIZETYPE_INT64   '/ ( -9223372036854775808ll, 9223372036854775807ll, 9223372036854775807ull ), _
```
dkl: Although `-9223372036854775808ll` seems to work "as expected" here, it may still be better to write it as `-9223372036854775807ll - 1` to avoid the constant overflow.

Also, how about writing these as hex to make typos easier to spot?

src/compiler/parser-decl-symbtype.bas
```diff
@@ -29,6 +29,85 @@ function cConstIntExpr _
...
+		errReport( FB_ERRMSG_EXPECTEDCONST )
+		astDelTree( expr )
+		expr = astNewCONSTi( 0, FB_DATATYPE_LONGINT )
+	end if
+
+	dtype = astGetDataType( expr )
+
+	value = astConstFlushToInt( expr, FB_DATATYPE_LONGINT )
+
+	if( typeIsSigned( dtype ) ) then
```
dkl: This check would be useful to have as a separate function, such as `isInRange` (or perhaps a `RangeValues.contains()` method), then maybe the `result` variable becomes unnecessary. (it seems to be an extension to astCheckConst())

src/compiler/parser-decl-symbtype.bas
```diff
@@ -29,6 +29,85 @@ function cConstIntExpr _
...
+		if( typeIsSigned( todtype ) ) then
+			result = (value >= r->smin) and (value <= r->smax)
+		else
+			if( dtype = FB_DATATYPE_LONGINT and todtype = FB_DATATYPE_ULONGINT ) then
+				result = (value >= 0) and (culngint(value) <= culngint(r->smax))
+			else
+				result = (value >= 0) and (culngint(value) <= culngint(r->umax))
```
dkl: umax doesn't need culngint() since it's already unsigned

src/compiler/parser-decl-symbtype.bas
```diff
@@ -29,6 +29,85 @@ function cConstIntExpr _
...
+
+	dtype = astGetDataType( expr )
+
+	value = astConstFlushToInt( expr, FB_DATATYPE_LONGINT )
+
+	if( typeIsSigned( dtype ) ) then
+		if( typeIsSigned( todtype ) ) then
+			result = (value >= r->smin) and (value <= r->smax)
+		else
+			if( dtype = FB_DATATYPE_LONGINT and todtype = FB_DATATYPE_ULONGINT ) then
```
dkl: may have to check sizetype instead of dtype here, since Integer can be 64bit too

src/compiler/parser-decl-symbtype.bas
jayrm: That makes sense.  I changed the table values to hex notation.  I also considered moving the table to symb-data.bas; there are a number of places in the compiler where these range limit values are used.  Or, maybe add a "fbc.bi" or "fb-limits.bi", either in the src/compiler directory, or in the ./inc directory and document in wiki.  Similiar to limits.h.

src/compiler/parser-decl-symbtype.bas
jayrm: Yea, not needed.  Removed.

src/compiler/parser-decl-symbtype.bas
jayrm: I changed to check by sizetype.  I realize I didn't have any INTEGER checks in the warning tests, so I added those too.

src/compiler/parser-decl-symbtype.bas
jayrm: I noticed that too; astCheckConst() is similar but less strict about overflows.  I will explore this a little more before closing off the pull request.

dkl:
Looks good!

jayrm:
In this last update, I have moved the range checking logic to it's own function hIsConstInRange() plus a few notes directly in the code, based on our discussion here, if anyone wants to expand on it.

PR #83, opened 2018-05-21
#879: Literal type suffix and unary operators

jayrm:
- preserve data type on numeric literals if used with unary operators and if literal is suffixed

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@ Version 1.06.0
 [changed]
 - Adjusted warning text for "mixed bool/nonbool operands" warning
 - test-suite uses libfbcunit for unit testing framework
+- #879: Literal type suffix and unary ops; preserve data type on numeric literals if used with unary operator and if literal is suffixed
 
 [added]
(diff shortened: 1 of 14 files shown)


dkl:
Should this behaviour be changed? It would make `-1l` behave differently than `-clng(1)` which seems wrong to me.

jayrm:
No, it shouldn't be changed,  Agreed, I think this change is worse than the reported bug.  Another example, if have `#define D = 1&` and `const C = 1&` then '-D' and '-C' will potentially be different data types.

Current fbc behaviour is consistent and can be explained easily.  In hind-sight, the real bug is that the documentation should be clearer.

Purpose of this pull request is to show what the change would look like, and bring the discussion at https://www.freebasic.net/forum/viewtopic.php?t=26629 and bug report at https://sourceforge.net/p/fbc/bugs/879/ justifiably to a close.

jayrm:
Proposed pull request closed. The proposed change causes more inconsistencies than it solves. Documentation updated.

PR #85, opened 2018-05-25
rtlib: fix some minor bugs in fb_CpuDetect for x86

tkchia:


jayrm:
What system/platform were you running on to notice this problem?
Thanks

tkchia:
Hello @jayrm, hello @dkl,

Actually I found the issues by eye, while trying to see if I could port FreeBASIC to a 16-bit platform.

The `cpuid` handling issue shows up (under a debugger) when I run the DJGPP port of `fbc` under [FreeDOS in QEMU](https://en.wikibooks.org/wiki/QEMU/FreeDOS). Under QEMU, `cpuid` with `eax == 1` gives `eax` (family/model/stepping) as `0x663`, and `edx` (feature flags) as `0x0781abfd`. This results in `detected_cpu` being set to `0x67b1abfd` (with two feature bits being set when they should not be).

Thank you!

jayrm:
Good eye!  Thanks for the fix.

PR #87, opened 2018-06-18
Fix dim shared byref initializer check

dkl:
global ref initializers were not checked for being non-constant, but that can actually matter. For example when trying to initialize a ref from another ref, the second ref will try to copy the first ref's value, which is a non-constant initializer that isn't possible without generating a global constructor.

So for now I think the best solution is to disallow the case - at least fbc doesn't crash anymore.

### DIFF
--- a/src/compiler/ast-node-typeini.bas
+++ b/src/compiler/ast-node-typeini.bas
@@ -556,7 +556,6 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL 
 		end if
 	'' literal string..
 	else
-		var sdtype = symbGetType( sym )
 		'' not a wstring?
 		if( sdtype <> FB_DATATYPE_WCHAR ) then
 			'' convert?
--- a/src/compiler/ast-node-typeini.bas
+++ b/src/compiler/ast-node-typeini.bas
@@ -524,7 +524,7 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL 
 	end if
 
 	var sdtype = symbGetType( sym )
-	var sfulldtype = symbGetType( sym )
+	var sfulldtype = symbGetFullType( sym )
 	if( symbIsRef( sym ) ) then
 		'' Initializers for references initialize the pointer,
 		'' not an object of the symbol's type.
--- a/src/compiler/ast-node-typeini.bas
+++ b/src/compiler/ast-node-typeini.bas
@@ -522,7 +522,15 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL 
 	if( sym = NULL ) then
 		sym = basesym
 	end if
+
 	var sdtype = symbGetType( sym )
+	var sfulldtype = symbGetType( sym )
+	if( symbIsRef( sym ) ) then
+		'' Initializers for references initialize the pointer,
+		'' not an object of the symbol's type.
+		sdtype = typeAddrOf( sdtype )
+		sfulldtype = typeAddrOf( sfulldtype )
+	end if
 
 	'' Get rhs expression
 	var expr = n->l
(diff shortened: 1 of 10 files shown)


jayrm:
Looks good to deal with the const initializer checks.

My only recommendation is, that after this PR is merged, is to close 814, 822, 842, and start a new ticket for the following:

bug ticket 822 has this example:
```    
Type UDT1
    Dim As Integer I1
End Type

Type UDT2
    Dim As Integer I2
    Static Byref As UDT1 ru1
End Type

Dim Byref As UDT1 UDT2.ru1 = *Cast(UDT1 Ptr, 0)
```
So we can still arrive at a failed assertion `assert( expr <> NULL )` hFlushExprStatic(), but I think this is a separate issue from the initial 822 bug report, and it would be easier to track as separate issue.

dkl:
Ok, that one is also interesting. Turns out the CONST code path (as opposed to the OFFSET code path) in hFlushExprStatic() didn't consider that a ref actually is a pointer and tried to do astNewCONV() from the type-casted null pointer initree (pointer type) to the symbol type (struct). The new commit should fix that.

jayrm:
You make it look so easy.  I struggled with this for days.  You are at about the spot I got stuck trying to convert the LHS sym to just a pointer.  x86 emitter doesn't know what to do with FB_DATATYPE_STRUCT in EMITVARINII.

jayrm:
In ten years I have never seen ".invalid" come up in x86 -gen gas.  Could use dtype = FB_DATATYPE_STRUCT to mean reference to struct. i.e. only time would ever pass STRUCT is that it is a ref to STRUCT.  Maybe.

In emit_x86.bas:EMITVARINII()
```
	case FB_DATATYPE_STRUCT, FB_DATATYPE_POINTER
		'' reference to struct or pointer
		function = @".long"
	case FB_DATATYPE_STRING
		function = @".INVALID"
```

Will allow this...
```
type udt1
	dim as integer i1
end type

type udt2
	dim as integer i2
	static byref as udt1 ru1
end type

dim byref as udt1 udt2.ru1 = *cast(udt1 ptr, 55)

print @udt2.ru1 '' returns 55
```

to succeed.

src/compiler/ast-node-typeini.bas
```diff
@@ -522,7 +522,15 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL
 	if( sym = NULL ) then
 		sym = basesym
 	end if
+
 	var sdtype = symbGetType( sym )
+	var sfulldtype = symbGetType( sym )
```
jayrm: `sfulldtype` has same value as `sdtype` throughout.

src/compiler/ast-node-typeini.bas
```diff
@@ -522,7 +522,15 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL
...
+	if( symbIsRef( sym ) ) then
+		'' Initializers for references initialize the pointer,
+		'' not an object of the symbol's type.
+		sdtype = typeAddrOf( sdtype )
```
jayrm: if this is removed....

src/compiler/ast-node-typeini.bas
```diff
@@ -543,7 +551,7 @@ private sub hFlushExprStatic( byval n as ASTNODE ptr, byval basesym as FBSYMBOL
 		else
 			'' different types?
 			if( edtype <> sdtype ) then
```
jayrm: ... then this should be `if( edtype <> sfluldtype ) then`

jayrm:
There is a lower level issue going on here in either x86 IR or x86 emitter.  -gen gcc seems to handle it this change OK, but -gen gas does not.  My previous comment is incorrect about replacing ".invalid" for structures in the x86 emitter.

For example, this succeeds for integers
```
Type UDT
	Dim As Integer I
	Static Byref As byte RI
End Type

Dim Byref As byte UDT.RI = *Cast(byte Ptr, 123456)

print @UDT.RI
```
but will give other ASM errors when emitted.

src/compiler/ast-node-typeini.bas
dkl: oops; fixed it.

PR #89, opened 2018-06-20
Add Byref Initializers of Byref Variables

jayrm:
- added hResolveRefToRefInitializer() called from hCheckAndBuildByrefInitializer()
- allows initializing a reference variable by referring to another reference variable.
- was discussed briefly in forum topic [Dim Byref Syntax](https://www.freebasic.net/forum/viewtopic.php?f=17&t=23584&start=60#p248549)


src/compiler/emit_x86.bas
```diff
@@ -4153,8 +4153,30 @@ private sub hCMPI _
 		ostr = "test " + dst + COMMA + dst
 		outp ostr
 	else
-		ostr = "cmp " + dst + COMMA + src
-		outp ostr
+
+		'' immediate and offset? it's invalid for CMP so use a temp reg
```
dkl: Somehow this issue seemed familiar - cc8638c2febc43c86e9a7994a87a4db349728ba6

src/compiler/emit_x86.bas
```diff
@@ -4153,8 +4153,30 @@ private sub hCMPI _
...
+		if( (svreg->typ = IR_VREGTYPE_IMM) and (dvreg->typ = IR_VREGTYPE_OFS) ) then
+
+			dim reg as integer, isfree as integer, rname as string
```
dkl: inner `rname` is shadowing the outer `rname` here, might be better to split into an extra function

src/compiler/parser-decl-var.bas
```diff
@@ -854,6 +854,49 @@ private function hBuildFakeByrefInitExpr( byval dtype as integer, byval subtype
...
+	dim tree as ASTNODE ptr = any
+	dim n as ASTNODE ptr = any
+
+	'' if initializer expr is a REF VAR, then crawl the INITREE of the var
+	'' it references for a VAR initializer.  If none found just return the 
+	'' original expr.
+
+	if( expr andalso astIsDEREF( expr ) ) then
+		if( expr->l andalso astIsVAR( expr->l ) ) then
+			if( expr->l->sym andalso symbIsRef( expr->l->sym ) ) then
```
dkl: It could be helpful to also check `symbIsVar`, since the `symbIsRef` flag can apply to other symbol types too, i.e. procedures for "return byref", and looking at `astNewVAR()` it refers to the procedure symbol in case of function pointers.

tests/dim/byref-shared-from-shared-ref.bas
```diff
@@ -1,4 +1,4 @@
-' TEST_MODE : COMPILE_ONLY_FAIL
+' TEST_MODE : COMPILE_ONLY_OK
```
dkl: we can probably remove this now, since the scenario is better covered by the new unit test

src/compiler/emit_x86.bas
jayrm: That's perfect.  Almost the same problem.  Will revert emit_x86.bas and let IR handle it.

src/compiler/parser-decl-var.bas
jayrm: Added symbIsVar check.  Thanks.  I think for the correct sym.

tests/dim/byref-shared-from-shared-ref.bas
jayrm: done

Issue #26, opened 2016-03-18
Parser allows more constructs under -exx than without it


StringEpsilon:
Duplicated here for BountySource.

Original Link: 
https://sourceforge.net/p/fbc/bugs/801/

cVarOrDeref() allows DEREFs but not CALLCTOR/TYPEINI nodes, so it behaves differently with/without -exx for expressions like *@type<...>(...) (with -exx the DEREF/ADDROF is preserved due to null pointer checks, but without -exx it is solved out).
-exx should not have this kind of effect on the parsing.
Maybe cVarOrDeref() should just accept CALLCTOR/TYPEINI nodes too, but then code like this would be accepted too:

```
type UDT
    i as integer
end type

dim x as UDT

type<UDT>(1) = x
```

which I don't think was ever intended to work. On the other hand, this already works:

```
type<UDT>().i = 1
```

so it wouldn't matter much.

StringEpsilon:
[![Bountysource](https://api.bountysource.com/badge/issue?issue_id=32035073)](https://www.bountysource.com/issues/32035073-parser-allows-more-constructs-under-exx-than-without-it?utm_source=32035073&utm_medium=shield&utm_campaign=ISSUE_BADGE)

jayrm:
Fixed at revision 391527e2a9421d54ae1b32c37259bcef4249f13f.  Requires fb.net bug fix https://sourceforge.net/p/fbc/bugs/886, and adds and additional code to solve out null pointer where not needed.

PR #90, opened 2018-06-26
Casting, Conversions, Const Qualifiers, Byref

jayrm:
For:
#642 Pointer casting allows constness dropping
#801 parser allows more constructs under -exx than without it 
#842 BYREF (variable) does not report some assignment errors detectable on its internal pointer 
#886 Inconsistent error with BYREF and CAST on CONST variables 

Primary issue is in AST, the conversion node (AST_NODECLASS_CONV), records information used later in code generation (i.e. different sizes or classes) and type information (i.e. different levels of pointer indirection or const-ness).  Often, the TYPE information needs to be preserved until an expression is completely parsed.  Because some type information is only present in the AST, it needs to remain valid at least long enough to be used.  In this patch, a separate flag is used to track const qualifier changes.

Secondary issue, is with -exx and Null ptr checks.  For some expressions, it's only after the expression is fully parsed can it be determined that any null pointer checks that were injected, should be removed.  Some null ptr checks were not being cleanly removed and so the resulting AST was different depending on if -exx was enabled or not.

The beginning of this pull request is a starting point for a WIP, and creating PR seems like an efficient way to develop and discuss the progress.

Some behaviour is only seen when '-w constness' command line option given, otherwise, most of the tests/warnings/test.sh tests will fail.  For example:
```
dim x as const integer = 1
dim p as integer ptr

p = cast( byte ptr, @x )

'' a.bas(4) warning 40(0): CONST qualifier discarded
'' a.bas(4) warning 4(1): Suspicious pointer assignment
```

There are several TODO's in the code to record next steps, or places the need more investigation.

The location of the new warning is not ideal.  Ideally, astNewCONV would just initialize the conversion node and the parser would already have determined if the operation is valid.  That's not always possible (or at least very complex/duplicating), and using astNewCONV to catch all and generate the warning is convenient.


src/compiler/ast-node-ptr.bas
```diff
@@ -36,18 +36,44 @@ function astNewDEREF _
...
 			case AST_NODECLASS_OFFSET
 				delchild = (t->ofs.ofs = 0)
+			
+			case AST_NODECLASS_PTRCHK
+
+				'' convert *PTRCHK(@expr) to (expr)
+				'' TODO: remove null-ptr checks in ptr indexing
+
+				if( (t->l->class = AST_NODECLASS_ADDROF) or _
+					(t->l->class = AST_NODECLASS_OFFSET and t->ofs.ofs = 0) ) then
```
dkl: `t->l->ofs.ofs = 0`?

StringEpsilon:
> #801 parser allows more constructs under -exx than without it

FYI: There is a 15 dollar bounty on that bug.

https://github.com/freebasic/fbc/issues/26

src/compiler/ast-node-ptr.bas
jayrm: oops! copy/edit typo

jayrm:
@StringEpsilon, I'll leave it up to you.  I'm satisfied just playing around, doing this for fun.

src/compiler/ast-node-conv.bas
```diff
@@ -472,6 +484,41 @@ function astNewCONV _
 	n->l = l
 	n->cast.doconv = doconv
 	n->cast.do_convfd2fs = FALSE
+	n->cast.convconst = FALSE
+
+	'' Discarding/changing const qualifier bits ?
+	'' TODO: reuse symbCheckConstAssign()?
```
dkl: yea, I think re-using symbCheckConstAssign and astPtrCheck (see astNewASSIGN/hCheckConstAndPointerOps) would be useful. The code below seems to be effectively the same as symbCheckConstAssign, and astPtrCheck uses the overload resolution (typeCalcMatch()) which will also check procedure pointer parameter types recursively.

```
sub sub_const(byref i as const integer)
	print i
end sub

sub sub_noconst(byref i as integer)
	i = 123
	print i
end sub

type type_const as sub(byref as const integer)
type type_noconst as sub(byref as integer)

dim ptr_const as type_const
dim ptr_noconst as type_noconst

ptr_const   = @sub_const   '' safe, same type
ptr_const   = @sub_noconst '' unsafe, because when calling through the ptr the param appears const, but the sub actually modifies it
ptr_noconst = @sub_const   '' safe, ptr allows more than the sub will do
ptr_noconst = @sub_noconst '' safe, same type

print cptr(sub(byref as const integer), @sub_const) '' safe
print cptr(sub(byref as const integer), @sub_noconst) '' unsafe, currently no warning even with -w constness
print cptr(sub(byref as integer), @sub_const) '' safe
print cptr(sub(byref as integer), @sub_noconst) '' safe
```

jayrm: I see.  Thanks.  So, need to go deeper here.  I will study to reuse the existing checking functions.  I have only been looking for places where might be manipulating the ast type directly.

A few months ago with https://sourceforge.net/p/fbc/bugs/880/ I was looking at typeCalcMatch() and related symb-proc.bas functions.  I started experimenting with:
* scoring the same const vs different const matches separately by increasing the range of the score 0 to 4*FB_DATATYPES
* counting the exact_matches and the const_matches separately
* making AST_OPOPT_NOCOERCION as 2 flags, one for left hand side, one for right hand side (though, I don't know if this helped and at the moment I can't remember why I was trying this)
I abandoned that effort, realizing I needed to study the code paths a lot more.  I dunno, I don't want to make this change too difficult, but maybe worth connecting the two ends of this experiment/effort.

jayrm:
I see the test build failed on compiling, CHttpStream.bas.  On fbc-win32/64 I get 2 warnings, but it doesn't hang.  I'll try on linux.

Still working though the PTR checking functions.  Intent of this update was to provide a some reference for the next changes.

* const checks are now enabled by default
* changed meaning of '-w constness', it overrides AST_CONVOPT_DONTWARNCONST, and enables all warnings to be shown, even the internal ones we would normally hide
* quiet warnings on some internal expressions
* don't warn for implicit cast to [const] zstring const ptr ptr in astBuildStrPtr()
* don't warn for internal expressions from astBuildDerefAddrOf()
* don't warn in astNewADDROF() if skipping nodes
* don't warn in hBuildByrefArg() because astNewARG() should have checked
* the intial test tests/warnings/const-discard.bas has some cases that need to be fixed/investigated

jayrm:
Now Tavis-CI succeeds.  In the last update, only change made was to revert when CONST Qualifier discarded warnings are printed.  I added the '-w constness' option to tests/warnings/test.sh.  But nowhere else.  What is interesting (annoying) is that Travic-CI now succeeds when compiling doc/libfbdoc, which leads me to believe that Travis-CI is failing, only because we are printing (outputting) a warning when making libfbdoc, and it is Travis-CI itself that is the real issue with redirected output; possibly hanging because previous command was a 'git' command and Travis-CI is suppressing output, preventing, what it might anticipate as a security problem.  This is all speculation of course, but otherwise relevant.

dkl:
Oh, maybe we need to add `</dev/null` (stdin redirection) to the build commands for fbdoc. The same was already needed for the other build commands (all that invoke fbc), because otherwise fbc would hang. See e.g. https://github.com/freebasic/fbc/blob/2f7d0047543bef49156a3dfd99be7355c66447d5/contrib/travis/build-and-test.sh#L54

It's due to FB rtlib doing a blocking read() from stdin during startup, trying to query some information from the terminal, or something like that. But sometimes there is no terminal. I'm not sure why it's not handled, afterall the rtlib checks isatty() in some places, maybe it's just missing somewhere.

dkl:
master now has 145c13f8ad41af71856b7efcebd13e927ca009ed which should fix the travis build issue.

jayrm:
dkl, thanks for the fix-up on Travis-CI.  

This next update doesn't have much new.  The changes were to help me follow some of the code, what I thought was a little more readable.
* Added #pragma constness to toggle the warning within the same source
* use named constants for everything to do with typeCalcMatch()

I am finding the code in and around astPtrCheck, astNewASSIGN, hCheckConstAndPointerOps a little tricky to follow.  I started creating directed graphs to help visualize what's going on.  Here's one of them [fbc-conv-checks.png 380K](http://www.execulink.com/~coder/tmp/fbc-conv-checks.png).  I feel like I'm close to a next step, but no breakthroughs yet.

dkl:
Same for me - I don't fully remember what these functions do and don't do. To be honest it actually is rather complicated since the language has lots of special cases with different behaviors in cast vs assign vs arg vs initializer, add to that overloading and byval/byref...

Ideally the compiler sources would be refactored to allow unit testing individual functions like these, maybe then it would be easier to understand.

jayrm:
Yeah, need to work through it line by line; I guess I'm being impatient.  The next update to this PR might just be comments and a few name changes...

I understand what you are saying, for several astNew*() procedures, the common problem to solve is to check every data type against every other data type, sub-types, plus all the options too, to determine action needed or failure.  The type checking procedures group some common scenarios, like same type & subtype, both ptr, both integer class, overloads, etc, going through the type checking matrix, but each astNew*() is little bit different.

There are 3 things that I am finding that I really have to focus on when following the code:
1) No comments and I'm not familiar enough yet with what each procedure is going to do.  The procedure name alone doesn't tell me enough.
2) symb*() procs leading to AstCheckConvNonPtrToPtr.  Maybe that's the only one.  That trips me up every time, because we have some symb*() procedures carrying an ASTNODE ptr argument, I lose track of which API I am in, AST or SYMB.
3) private h*() helper procs, that have neither AST or SYMB prefixes calling out to other public AST/SYMB.  Have to pay attention to what file I am in.

---
I have had thoughts about internal testing, wanting to call a procedure with every possible value directly; here's some (maybe crazy) ideas that are probably a lot of work:

1) Instead of fbc's usual main(), invoke an internal test-suite.  But, without an input source file, tests would need to be fairly low level, constructing any needed global data manually.

2) new directives, especially for the compile_only_ok/fail tests to allow grouping tests in to one file, by counting errors/warnings, and resetting the compile state for each segment.
```
#__fb_test__ "compile_only_ok"
var a = 1

#__fb_test__ "compile_only_ok"
var a = 1

#__fb_test__ "compile_only_fail"
var a = 1
var a = 1
```
3) special build of fbc that allows input .bas source file to direct fbc to call an internal test procedure.  For example, assuming fbc was compiled with a test procedure "sub test_1( byval as ASTNODE ptr = NULL )" built-in, then the input source .bas could possibly be:
```
#__fb_test__ "call", "test_1"
'' or
var a=1, b=1.0, c=__fb_test__( "test_1", a + b )
```
4) Special build of fbc, to allow a second special build of fbc that registers hook procedure declarations, that allows the input .bas source file to select an alternate internal procedure call, which contains the test.  For example, instead of usual call to astNewBOP(), it is emitted as a function ptr look-up and call directed by the input test file.

in fbc:
```
#__fbc_test__ "register hook", "astNewBOP", "astNewBOP_test_1"
function astNewBOP_test_1( ... )
	'' tests before call
	...

	'' call original (don't hook here)
	function = astNewBop( ... )

	'' tests after call
	...
	
end function
```
input file for testing:
```
#__fbc_test__ "hook", "astNewBOP", "astNewBOP_test_1"
var a = 1, b = 1.0, c = a + b
```

jayrm:
At last, some progress.  This update is first attempt at a fix for bug 886.  I added symbCheckConstAssignFuncPtr() to do the function pointer checking.  Re-using symbCheckConstAssign() as the entry point was awkward, due the recursive nature of the type-checking procedures.  I think it could still be improved, and probably should need many more tests.

Next steps, in the current tests/warnings/const-discard.bas, the warnings on a few  memory related rtlib (like deallocate & clear).

jayrm:
Last update is rebase onto current master, to get rid of some of the weird merge history over last month.  End result is same except for changelog.txt.  For now, have kept the 2 revision histories in https://github.com/jayrm/fbc/tree/cast-conv-ptr and https://github.com/jayrm/fbc/tree/cast-conv-ptr-old

jayrm:
In this last update fde631ef2f6581efb0a5cbc658a90fecfa229668 one Travis build job failed due to network, can ignore, I think.

Fixed so far:
- sf.net #642 Pointer casting allows constness dropping
- sf.net #801 parser allows more constructs under -exx than without it
- sf.net #842 BYREF (variable) does not report some assignment errors detectable on its internal pointer
- sf.net #886 Inconsistent error with BYREF and CAST on CONST variables
- compiler recurses in to function pointer types to check compatibility, returning additional warning messages.

Notes on code changes so far:
- the CONST check warnings are reported only if '-w constness' command line option given.  Added this option to tests/warnings/test.sh, so there are a few existing warning tests that get additional warnings generated. 

- all changes for these bugs are in the form of compiler warnings only, so this part of the update should not break any existing code.  

- About promoting CONST checks to errors: callers of astNewCONV may have checked for invalid conversions before trying to call astNewCONV, or maybe not.  So, astNewCONV is last opportunity to check for a valid conversion; making the CONSTness checks here is valid as they are presented as warnings.  To promote a CONST check to error, either have to, do the check before calling astNewCONV and present an error, or call astNewCONV first and examine the error/warning returned after.

- symbCheckConstAssign*() procs provide a boolean pass/fail result, but also find out other infrormation, and while similar, not exactly the same as symbCalcProcMatch().  Not sure where the balance should be, between many procedures performing specific tasks, or a few general purpose methods calculating lots of information that may or may not be needed.

WIP:
- #727 RTL functions not checking Constness unless marked RTL_CONST

- In this last update added a base test in tests/warnings/rtl-prototypes.bas to get function pointer of every built-in rtlib functin.  It will only warn if CONST was added to the test function pointer declaration but was never added to built-in RTL prototype (can't check the reverse though).

- To make this test work, had to allow WSTRING to be returned as a function result, but only if a prototype.  Otherwise, it is impossible to create a function pointer to a built-in rtlib function.  I think this is valid since it represents current state of the compiler.  Though some users may find it confusing that can declare WSTRING return, but not write a user function that returns WSTRING.

jayrm:
dkl, I requested a review because I think I am close to wrapping up this work.

I feel confident with about 90% of what I've done here.  If there was an issue reported, I feel like I could address it effeciently.

Naturally, any of the changes I have made are open to criticism.  However, I would like to direct to a few items specifically:

1) I think '-w constness' reports CONST discarded well.  Actually, when I run it on some of my own code having heavy pointer use, it's kind of annoying, because I know that I've written is what I want to do.  That said, it does show where I could possibly express my conversions differently.

```
#pragma constness=true
dim c as const integer ptr = 0
dim p as integer ptr = cast( integer ptr, c ) '' warning
```

2) rtl-*.bas changes.  Basically, I've added CONST to everything that could be CONST, even BYVAL parameters.  C name mangling doesn't involve CONST, and the interface between fbc and rtlib is valid.  Although CONST in fbc built-in delcaration but not in rtlib C declaration might be seen as confusing.

3) overload resolution changes in 208fad41acd3a0ee2d2aaae189b349c8e16fb897, this revision was somewhat rushed.  I am currently working on a generator for testing overload resolution in the form of:
```
sub S(param1)
sub S(param2)
call S(arg)
```
... so I can test what changed between fbc 1.0.5 and current master.  With a test generator, it results in 10's of thousands of tests, that each have to be complied separately.  It takes a long time to run.  I might be leaning towards good code review, and hoping for the best. :)  The commit message should also describe what I am trying to do here.

Anyway, if you have some free time, would appreciate if you could comment on your most critical concerns.

src/compiler/symb.bas
```diff
@@ -1920,9 +1925,14 @@ function symbCheckConstAssign _
 	function = FALSE
 	matches = 0
 
-	'' no consts? short-circuit
-	if( (typeGetConstMask( ldtype ) or typeGetConstMask( rdtype )) = 0 ) then
-		return TRUE
+	'' neither is a function pointer?
+	if( ( typeGetDtOnly( ldType ) <> FB_DATATYPE_FUNCTION ) and ( typeGetDtOnly( rdType ) <> FB_DATATYPE_FUNCTION ) ) then
```
dkl: Why not shortcut the case of function pointers without consts anymore? As far as I can see the code below only checks the function ptr dtype const bits, and not parameters on the subtype, anyways.

src/compiler/symb.bas
```diff
@@ -1977,6 +1987,124 @@ function symbCheckConstAssign _
...
+	errmsg = FB_ERRMSG_OK '' FB_ERRMSG
+	wrnmsg = 0            '' FB_WARNINGMSG
+
+	assert( typeGetDtOnly( ldType ) = FB_DATATYPE_FUNCTION )
+	assert( typeGetDtOnly( rdType ) = FB_DATATYPE_FUNCTION )
+	assert( symbIsProc( lsubtype ) )
+	assert( symbIsProc( rsubtype ) )
+
+	'' not a function pointer?
+	if( ( typeGetDtOnly( ldType ) <> FB_DATATYPE_FUNCTION ) or ( typeGetDtOnly( rdType ) <> FB_DATATYPE_FUNCTION ) ) then
```
dkl: (given the assert()s above, this may aswell be removed)

src/compiler/symb.bas
```diff
@@ -1977,6 +1987,124 @@ function symbCheckConstAssign _
 	function = TRUE
 end function
 
+function symbCheckConstAssignFuncPtr _
```
dkl: This function seems very similar to `symbCalcProcMatch()`, it would be nice to share some of the code. (I'm thinking of a generic function signature comparison/compatibility check)

src/compiler/symb.bi
```diff
@@ -1865,14 +1871,40 @@ declare function symbGetDefaultParamMode _
 
 declare function symbVarCheckAccess( byval sym as FBSYMBOL ptr ) as integer
 
+declare function symbCheckConstAssignTopLevel _
```
dkl: Do we even need the special case versions to be public anymore? Can't all code just use the new generic `symbCheckConstAssign()`?

src/compiler/symb.bi
```diff
@@ -2010,9 +2042,6 @@ declare function symbGetUDTBaseLevel _
 #define symbGetIsUnionField(s) ((s->stats and FB_SYMBSTATS_UNIONFIELD) <> 0)
 #define symbSetIsUnionField(s) s->stats or= FB_SYMBSTATS_UNIONFIELD
 
-#define symbGetIsRTLConst(s) ((s->stats and FB_SYMBSTATS_RTL_CONST) <> 0)
```
dkl: (good to see this special case removed...)

tests/const/rtl-clear.bas
```diff
@@ -0,0 +1,9 @@
+' TEST_MODE : COMPILE_ONLY_FAIL
+
+'' from sf.net #642
+
+scope
+
+	dim as const integer a = 256
+
+	clear a, 0, sizeof(integer)
```
dkl: Missing an `end scope` it seems.

tests/structs/anon-invalid-initializer.bas
```diff
@@ -0,0 +1,20 @@
...
+
+	dim x as UDT
+
+	'' OK
+   	type<UDT>(0).i = 1
+	*@type<UDT>(0).i = 1
+
+    '' Expected to fail
+	type<UDT>(1) = x
+	*@type<UDT>(1) = x
```
dkl: If both cases are expected to fail independently, then they should be separate test cases to check that.

src/compiler/rtl-array.bas
```diff
@@ -26,20 +26,20 @@
...
 			} _
 		), _
 		/' function fb_ArrayRedimPresvEx cdecl _
 			( _
 				array() as any, _
-				byval elementlen as uinteger, _
-				byval doclear as long, _
-				byval isvarlen as long, _
-				byval dimensions as uinteger, _
+				byval elementlen as const uinteger, _
```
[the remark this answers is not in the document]
jayrm: I really struggled with this decision exactly because of "byval" being treated same as "byval as const".  I don't know, I may have misguided logic here: for the caller, byval implies const, so it seems more descriptive to explicitly include const in the prototype.  Yet, fbc doesn't (or shouldn't) check for const on byval and C ignores it, so end result is useless.  I have no problem rolling back the changes and just doing the byref/ptr types; maybe that would be clearer having the fb rtl built-ins follow what is actually in the C rtlib public API.

src/compiler/symb-proc.bas
```diff
@@ -1674,23 +1676,23 @@ private function hCalcTypesDiff _
 			''   it's a DEREF or not (it can all be treated as string)
 			'' - disallow other args (passing BYVAL explicitly
 			''   should be handled by caller already)
-			select case( param_dtype )
+			select case( typeGetDtAndPtrOnly( param_dtype ) )
```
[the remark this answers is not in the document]
jayrm: agreed, this is leftovers from something I tried and didn't fully revert

src/compiler/symb-proc.bas
```diff
@@ -2388,13 +2378,20 @@ private function hCheckCastOvl _
 
 	'' same types?
 	if( typeGetDtAndPtrOnly( proc_dtype ) = typeGetDtAndPtrOnly( to_dtype ) ) then
+		
 		'' same subtype?
 		if( proc_subtype = to_subtype ) then
-			return FB_OVLPROC_FULLMATCH
+
+			'' same const?
+			if( proc_subtype = to_subtype ) then
```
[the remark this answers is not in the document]
jayrm: Yes, should be dtype.  Worries me a little that the current test-suite passes with and without this check for full match.  Either there is no specific test for what's going on here, or it doesn't matter.  Hopefully more testing will prove it out one way or the other.

src/compiler/symb.bas
jayrm: Yes, can short-cut, left over from previous work.

src/compiler/symb.bas
jayrm: ditto as above

src/compiler/symb.bas
jayrm: yea, that's been on my mind as well.  Maybe you can see from the changes I've made to the match score, I'm trying to head in that direction.  The idea is to give additional meaning to the match score.  Possibly even embedding information in the least significant bits.

src/compiler/symb.bi
jayrm: haha, I don't know.  After 4 weeks of looking at the same 12 procs everyday, my brain was in knots.  I did try symbCheckConstAssign() as the generic initial proc at some point in my efforts, but at the time I tried it, I had infinite recursion and couldn't sort out how to solve it.  That's why there was some leftover checks for function types in a couple procs.  I think I can make another attempt at solving this out... and not go insane.

jayrm:
In 64d32f5ca2239d09944f50e887bedd4b0fd147a2 some minor cleanups that I think make the code a little easier to follow.

- symbCheckConstAssign(), public, a generic entry point
- symbCheckConstAssignTopLevel(), public, check top level only
- hSymbCheckConstAssignFuncPtr(), private, check the func ptr sub-type only

So, 
- no recursion nightmare like I had before
- can use the generic symbCheckConstAssign() anywhere currently using symbCheckConstAssignTopLevel()
- could combine these procs more if desired, probably just to a single (recursive) procedure. 
- or if they stay separate, there are unused parameters that could be removed 

**But first,** should compare what's going to change.  Consider astNewARG() using this test code:
```
declare sub byref_ci__( byval as sub( byref as const integer ) )
dim _i__ as sub( byref as integer )
byref_ci__( _i__ ) '' unsafe
```
Currently, 
- in astNewARG(), we call symbCheckConstAssignTopLevel() only, and with the top-level const OK, but pointer types different, we get a warning only:
`Passing different pointer types, at parameter 1 of BYREF_CI__()`
- And, we preserve constness, so later in astNewCONV() when we do the full const check, we also get a warning only:
`CONST qualifier discarded, at parameter 1 of BYREF_CI__()`

Now compare, 
- if we instead call symbCheckConstAssign() in astNewARG() to do the full check on both the direct assignment and the function pointer type, we get an error, instead:
`Invalid assignment/conversion, at parameter 1 of BYREF_CI__() in 'byref_ci__( _i__ )'`

Maybe an error is in fact what we want here, but, **I'm a little nervous** about the scope of this change doing it this way, though, should only affect function pointers.

src/compiler/symb.bas
```diff
@@ -1885,9 +1885,6 @@ function symbCheckConstAssignTopLevel _
 	dim as integer i = any, lcount = any, rcount = any, rmatches = any
 	dim as integer lconst = any, rconst = any
 
-	errmsg = 0
```
dkl: (looks like these parameters are unused, so they may aswell be removed)

src/compiler/symb.bas
```diff
@@ -1987,7 +1979,7 @@ function symbCheckConstAssignTopLevel _
 	function = TRUE
 end function
 
-function symbCheckConstAssignFuncPtr _
+function hSymbCheckConstAssignFuncPtr _
```
dkl: This function can be `private` now

jayrm:
I kept the l/rdtype l/rsubtype in the procedures, even though they are not all used currently, because I think eventually, if we can refactor to a common generic checking function, we will need them.  I tidied the changelog.txt file a little and recorded some notes in both the source code and the last commit message for this series of revisions.  Hopefully, it will be enough info to pick up the efforts later.

PR #94, opened 2018-08-31
cast-conv-ptr: don't warn on function pointer types by default

jayrm:
- added AST_CONVOPT_DONTWARNFUNCPTR to disable pointer warnings
- added '-w funcptr' to enable function pointer warnings in unary hCast()
- updated warning messages with "in function pointer" appended


jayrm:
For now, have added '-w funcptr' compiler option to enable to the function pointer checks.

If CAST()/CPTR() causes a warning, even optionally, like with '-w pedantic' then there needs to be a sensible way to formally express it another way.  Otherwise, there's no way to not get the warning.  Herein lies the problem.  The only way we have to express these explicit function pointer conversions is CAST()/CPTR().

src/compiler/symb.bas
```diff
@@ -2017,32 +2017,26 @@ private function hSymbCheckConstAssignFuncPtr _
...
+	elseif( symbIsRef( lsubtype ) <> symbIsRef( rsubtype ) ) then
 		wrnmsg = FB_WARNINGMSG_RETURNMETHODMISMATCH
-		exit function
-	end if
 
 	'' Different calling convention?
-	if( symbAreProcModesEqual( lsubtype, rsubtype ) = FALSE ) then
+	elseif( symbAreProcModesEqual( lsubtype, rsubtype ) = FALSE ) then
 		wrnmsg = FB_WARNINGMSG_CALLINGCONVMISMATCH
-		exit function
```
dkl: Wouldn't removing the early exits change the behaviour, i.e. it returns TRUE now (as long as the loop checking the parameters doesn't detect any problem)?

src/compiler/ast-node-conv.bas
```diff
@@ -492,22 +492,27 @@ function astNewCONV _
 
 		n->cast.convconst = ( symbCheckConstAssign( to_dtype, ldtype, to_subtype, l->subtype, , , wrnmsg ) = FALSE )
 
-		'' else check if const conversion
+		'' use the function pointer warning message?
+		if( wrnmsg <> 0 ) then
+			if( (options and AST_CONVOPT_DONTWARNFUNCPTR) = 0 ) then
```
dkl: Doesn't the code here also handle non-function pointers? So I'm not fully sure which part ensures that the new option affects function pointers only.

src/compiler/symb.bas
jayrm: Yea.  The intent was return both parts separately:
TRUE/FALSE - the consts match or not
wrnmsg=0, <>0 - the function pointer args, callconv, etc, match or not

I don't think this quite right.  Doen't make much sense to checks consts, if the function ptr is fundamentally mismatched.

src/compiler/ast-node-conv.bas
jayrm: True, handles all pointers.  It's a hack: currently the warning message is only set by hSymbCheckConstAssignFuncPtr() which is only called for function pointer types.

jayrm:
Maybe this would work:

Assuming that we've reverted to having early out for function pointer mismatches
and we have -w funcptr & -w constness

In all cases, **start with the full check**, the
1) Neither option: don't warn anything
2) Both -w funcptr & -w constness: warn everything
3) -w funcptr only: warn if wrnmsg<>0
4) -w constness only: this one's trickier; warn if wrnmsg<>0 AND at least one type (or subtype) has a const bit set.

src/compiler/ast-node-conv.bas
```diff
@@ -492,22 +492,36 @@ function astNewCONV _
...
 		if( n->cast.convconst ) then
-			if( (options and AST_CONVOPT_DONTWARNCONST) = 0 ) then
-				if( fbPdCheckIsSet( FB_PDCHECK_CONSTNESS ) ) then
-					'' specific warning message takes priority over const warning
-					if( wrnmsg = 0 ) then
-						wrnmsg = FB_WARNINGMSG_CONSTQUALIFIERDISCARDED
+
+			'' wrnmsg is <> 0 only if funcptr check failed
+			'' specific warning message takes priority over const warning
+			if( wrnmsg <> 0 ) then
```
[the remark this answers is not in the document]
dkl: ah ok, I missed that. Maybe it should return a result struct containing the various information (constness compatibility, procedure pointer compatibility, warning message)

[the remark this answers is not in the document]
jayrm: Yea, a struct would be better.  I imagine that if we were to have a have a generic type checking function, we would need something like anyway.  I'm not sure what form that is all going to take, so  I'll leave it as-is for right now.

jayrm:
I made a small change to the warning levels so that default compile '-w 0' and '-w pedantic' would generate same quantity of warnings as fb 1.05, essentially hiding the new function pointer warnings based on level.  Interesting that '-w all' is actually equivalent for '-w -1' since I think first versions.  I'm not sure what was intended there because it looks a little weird when the warning is printed out with level (-1).  We don't have any (-1) level warning messages so never would have noticed it before.

Originally I was going to bump all the error messages up by one and make the default warning level 1 which seems like the right thing to do.  Instead, I decided that we need to have this version first, to show why we would change the '-w level' behaviour.

PR #98, opened 2018-09-13
Added  __FB_GUI__ intrinsic define

marpon:
only 1 define as console is default mode, compiled ok with 32 and 64

tested with
#if __FB_GUI__
#include "windows.bi"
messagebox(0, "Gui mode", "info",0) 
#else
print "Console mode"
print : print "Any key to close"
sleep
#endif

### DIFF
--- a/src/compiler/symb-define.bas
+++ b/src/compiler/symb-define.bas
@@ -85,7 +85,11 @@ end function
 private function hDefOptGosub_cb ( ) as string
 	function = str( env.opt.gosub = TRUE )
 end function
-
+															
+private function hDefViewGui_cb ( ) as string            '***** console/gui
+	function = str( env.clopt.modeview = FB_MODEVIEW_ISGUI )
+end function
+																
 private function hDefOutExe_cb ( ) as string
 	function = str( env.clopt.outtype = FB_OUTTYPE_EXECUTABLE )
 end function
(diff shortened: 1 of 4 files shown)


dkl:
Thanks for the patch, looks good. I only have some nitpicking comments about formatting:
- Could you remove the `'***** console/gui` comments? This is generally not done in the fbc source code.
- Could you fix the indentation of the added `select case` block; currently it seems to have start/end at different indentation levels.
- As a tip for the future: You can put modifications to multiple files in a single commit, which makes reviewing now and later a bit easier.

src/compiler/symb-define.bas
```diff
@@ -85,7 +85,11 @@ end function
 private function hDefOptGosub_cb ( ) as string
 	function = str( env.opt.gosub = TRUE )
 end function
-
+															
+private function hDefViewGui_cb ( ) as string            '***** console/gui
+	function = str( env.clopt.modeview = FB_MODEVIEW_ISGUI )
```
jayrm: Is a new option needed?  This call back could check the value of fbc.subsystem for "gui" indicator.

marpon: please clear that request, i'm sendind a cleaned right now
thanks for supporting my wrong attempt

PR #101, opened 2018-09-14
Cleaned request  __FB_GUI__ intrinsic define

marpon:
Here you can find my last (cleaner) version, better to close the previous one.

only 1 define as console is default mode, compiled ok with 32 and 64
tested with
#if FB_GUI
#include "windows.bi"
messagebox(0, "Gui mode", "info",0)
#else
print "Console mode"
print : print "Any key to close"
sleep
#endif

### DIFF
--- a/src/compiler/fb.bas
+++ b/src/compiler/fb.bas
@@ -480,7 +480,8 @@ sub fbGlobalInit()
 	env.clopt.stacksize     = FB_DEFSTACKSIZE
 	env.clopt.objinfo       = TRUE
 	env.clopt.showincludes  = FALSE
-
+	env.clopt.modeview      = FB_DEFAULT_MODEVIEW
+				
 	hUpdateLangOptions( )
 	hUpdateTargetOptions( )
 end sub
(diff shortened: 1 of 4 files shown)


marpon:
Hope this time it is better organized...

for the sleep evolution, it still quite difficult to trap it on the compiler bas files, i think it has to be done on the rtl  c code files.

I will try , but 1 question wich version of mingw gcc do you promote to compile the c parts?

jayrm:
marpon, there are several tool chains available that can work on windows.

Currently I use mingw-w64 versions:
gcc (i686-win32-sjlj-rev1, Built by MinGW-W64 project) 7.1.0
gcc (x86_64-win32-sjlj-rev1, Built by MinGW-W64 project) 7.1.0

And sometimes old mingw (32bit only) version: gcc (GCC) 4.8.1

jayrm:
oops! clicked the wrong buttons there.  The last check was good though.

marpon:
Thank's for your very fast action, and for the info about tool-chain 
i also use the MinGW-W64 7.1.0 for 32/64  and sometimes the MinGW-W64 5.3.0

i'm investigating on how to know on fb_sleep so in fact in rtlib  
if gfx screen is in use
and / or  if  console is in use  
because , even on gui mode in windows it is always possible to attach a console

and after to throw the compile error according the situation.
it will probably wait a while...
today outside compiler with __FB_GUI__  it is possible to manage better.

PR #105, opened 2018-09-25
Name mangling on win target

jayrm:
[#890](https://sourceforge.net/p/fbc/bugs/890/): Fix Bad @N stdcall suffix for non-trivial byval type parameters with -gen gcc

Changes in this pull request are based on, and related to:
[#828](https://sourceforge.net/p/fbc/bugs/828/) 32bit Integer/Long mangling reversed

By default, fbc's Long/ULong data type maps to c's int type, and since Integer/Uinteger are meant to be (consistently) 64bits in fbc-64bit, there is no other type that (by default) that can map to c's long int, which is 32bit even on Win64.

So, in Win64, to allow calling an external library that needs a 'long int' parameter there is the following, non-portable mangling modifier.  It is ignored for all targets except Win64. 

map fbc's 32bit 'LONG' type to Win64's 32bit 'long' type
`type cxxlongint as long alias "long"`

### DIFF
--- a/src/compiler/ir-hlc.bas
+++ b/src/compiler/ir-hlc.bas
@@ -505,11 +505,13 @@ private function hNeedAlias( byval proc as FBSYMBOL ptr ) as integer
 		function = TRUE
 
 	'' For stdcall with @N suffix, if the function has a hidden UDT result
-	'' pointer parameter, we need the alias to get the correct @N suffix.
-	'' (gcc would calculate the parameter into the @N suffix, since it
-	'' doesn't know that the parameter is the special result parameter)
+	'' pointer parameter, or UDT's passed byval, we need the alias to get
+	'' the correct @N suffix. (gcc could calculate the wrong value into the
+	'' @N suffix, since it doesn't known about the special result parameter
+	'' or byval UDTs).  It should be safe to always generate the alias
+	'' ourselves since we already must control for the special cases.
 	case FB_FUNCMODE_STDCALL
-		function = symbProcReturnsOnStack( proc )
+		function = TRUE
 	end select
 end function
 
--- a/src/compiler/parser-decl-symbtype.bas
+++ b/src/compiler/parser-decl-symbtype.bas
@@ -393,6 +393,38 @@ function hIntegerTypeFromBitSize _
 
 end function
 
+private function cMangleModifier _
+	( _
+	) as integer
+
+	static as zstring * FB_MAXNAMELEN+1 aliasid
+	function = FALSE
+
+	'' ALIAS?
+	if( lexGetToken( ) = FB_TK_ALIAS ) then
+		lexSkipToken( )
+
+		'' "long"
+		if( lexGetClass( ) = FB_TKCLASS_STRLITERAL ) then
+			aliasid = *lexGetText( )
+			lexSkipToken( )
+
+			select case lcase( aliasid )
+			case "long"
+				function = fbIs64bit( ) and ((env.target.options and FB_TARGETOPT_UNIX) = 0)
+			case ""
+				errReport( FB_ERRMSG_EMPTYALIASSTRING )
+			case else
+				errReport( FB_ERRMSG_SYNTAXERROR )	
+			end select
+		else
+			errReport( FB_ERRMSG_SYNTAXERROR )
+		end if
+
+	end if
+
+end function
+
 '':::::
 ''SymbolType      =   CONST? UNSIGNED? (
 ''				      ANY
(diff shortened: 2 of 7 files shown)


src/compiler/ir-hlc.bas
```diff
@@ -505,11 +505,13 @@ private function hNeedAlias( byval proc as FBSYMBOL ptr ) as integer
...
-	'' (gcc would calculate the parameter into the @N suffix, since it
-	'' doesn't know that the parameter is the special result parameter)
+	'' pointer parameter, or UDT's passed byval, we need the alias to get
+	'' the correct @N suffix. (gcc could calculate the wrong value into the
+	'' @N suffix, since it doesn't known about the special result parameter
+	'' or byval UDTs).  It should be safe to always generate the alias
+	'' ourselves since we already must control for the special cases.
 	case FB_FUNCMODE_STDCALL
-		function = symbProcReturnsOnStack( proc )
+		function = TRUE
```
dkl: Agreed, I think that's the right fix for this.

src/compiler/symb.bi
```diff
@@ -2492,6 +2495,10 @@ declare sub symbProcRecalcRealType( byval proc as FBSYMBOL ptr )
 #define	typeSetIsRef( dt ) (dt or FB_DATATYPE_REFERENCE)
 #define	typeUnsetIsRef( dt ) (dt and not FB_DATATYPE_REFERENCE)
 
+#define	typeIsMangleDt( dt ) (((dt and FB_DT_MANGLEMASK) <> 0))
```
dkl: maybe `typeHasMangleDt` would be a better name, because `typeIsMangleDt` made me think it checks whether the given dtype is one that can be mangled.

src/compiler/parser-decl-symbtype.bas
```diff
@@ -393,6 +393,38 @@ function hIntegerTypeFromBitSize _
 
 end function
 
+private function cMangleModifier _
```
dkl: Interesting, I like this idea. It could even be useful for some other things, like mangling to match C++ __int128.

src/compiler/parser-decl-symbtype.bas
```diff
@@ -393,6 +393,38 @@ function hIntegerTypeFromBitSize _
...
+	static as zstring * FB_MAXNAMELEN+1 aliasid
+	function = FALSE
+
+	'' ALIAS?
+	if( lexGetToken( ) = FB_TK_ALIAS ) then
+		lexSkipToken( )
+
+		'' "long"
+		if( lexGetClass( ) = FB_TKCLASS_STRLITERAL ) then
+			aliasid = *lexGetText( )
```
dkl: The aliasid static var seems unnecessary, since it's only used locally - I think it would be good to eliminate it (by checking the string immediately after lexGetText() and moving the lexSkipToken() behind the check).

src/compiler/symb.bi
jayrm: > maybe `typeHasMangleDt` would be a better name, because `typeIsMangleDt` made me think it checks whether the given dtype is one that can be mangled.

Makes sense to me, OK.

src/compiler/parser-decl-symbtype.bas
jayrm: Cool.  I added a note to the source in cMangleModifier() to explain the choices so far and possible next steps.

> Interesting, I like this idea. It could even be useful for some other things, like mangling to match C++ __int128.

I just tried this little test on Win64 by allowing `alias "__int128"` to be attached to a struct (used an arbitrary mangle_dt number to map the mangling code 'o' for unsigned __int128):
```
unsigned long long int get128_lo( unsigned __int128 x )
{ return (unsigned long long int)x; }

unsigned long long int get128_hi( unsigned __int128 x )
{ return (unsigned long long int)(x >> 64); }
```
```
type uint128_t
	lo as ulongint
	hi as ulongint
end type

extern "c++"
	declare function get128_lo( byval x as uint128_t alias "__int128" ) as ulongint
	declare function get128_hi( byval x as uint128_t alias "__int128" ) as ulongint
end extern

dim x as uint128_t = ( &h12345678, &h87654321 )

print hex( get128_lo(x) )
print hex( get128_hi(x) )
```

And it worked!  Surely, I must be missing something... it's never that easy.

src/compiler/parser-decl-symbtype.bas
jayrm: Removed the static aliasid,  Yea, an earlier version was returning a string, but that turned out to be not very useful.  Thanks.

PR #106, opened 2018-09-25
GCC warnings

countingpine:
This is my first commit for a while.  Doing a PR seems like a nice way to go about this.
This fixes some compiler warnings I came across while making from Ubuntu 18.04 (64-bit).
Most are implicit `switch()` fall-throughs, two are `sprintf()` buffer allocation warnings.

Note that I've made one of the fall-throughs into a `goto`, to match the other cases in that block.  The fall-through worked because it was the last case in the block, but could have broken if another case were added.  The compiler should be able to optimise it out anyway.

The joystick `sprintf()` warning was difficult to make a decision about, because the problem is GCC not understanding that `j` can only be two digits.  I extended the buffer, but other approaches might involve changing the type of `j`, using character manipulation or something to write the digits, or just ignoring the warning.

With these changes, the rtlib and gfxlib compile without warnings in (my version of) Linux.

### DIFF
--- a/src/gfxlib2/gfx_draw.c
+++ b/src/gfxlib2/gfx_draw.c
@@ -220,9 +220,9 @@ FBCALL void fb_GfxDraw(void *target, FBSTRING *command)
 				move = draw = TRUE;
 				break;
 
-			case 'F': case 'D': angle += 90;
-			case 'G': case 'L': angle += 90;
-			case 'H': case 'U': angle += 90;
+			case 'F': case 'D': angle += 90; /* fall through */
+			case 'G': case 'L': angle += 90; /* fall through */
+			case 'H': case 'U': angle += 90; /* fall through */
 			case 'E': case 'R':
 				diagonal = ((toupper(*c) >= 'E') && (toupper(*c) <= 'H'));
 				c++;
--- a/src/gfxlib2/linux/gfx_joystick.c
+++ b/src/gfxlib2/linux/gfx_joystick.c
@@ -34,7 +34,8 @@ FBCALL int fb_GfxGetJoystick(int id, ssize_t *buttons, float *a1, float *a2, flo
 	const char *device[] = { "/dev/input/js",
 							 "/dev/js",
 							 NULL };
-	char device_name[16];
+	/* overallocate device_name[] to prevent a sprintf() warning in GCC */
+	char device_name[13 + 11 + 1];
 	JOYDATA *joy;
 	JS_EVENT event;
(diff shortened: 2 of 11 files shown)


jayrm:
Nice!  

I compiled for win32/64 and see fall through warnings for following lines.
```
src/rtlib/dev_file_encod_read_core.c: 168, 171, 174, 177, 180
src/rtlib/utf_convfrom_wchar.c: 39, 42, 45, 80, 83, 86
src/rtlib/utf_convto_char.c: 28, 30, 32, 34, 36, 75, 77, 79, 81, 83
src/rtlib/utf_convto_wchar.c: 32, 34, 36, 38, 40, 90, 92, 94, 96, 98
src/rtlib/utf_convto_wchar.c: 148, 150, 152, 154, 156, 192, 194, 196, 198, 200
src/gfxlib2/win32/gfx_win32.c: 329
```
I didn't try dos (my cross compiler not working).

countingpine:
Not sure how I missed some of them.  Surely GCC would have caught those ones too.. I'll have another look.

The gfxlib code seems quite complex.  If it is genuinely supposed to fall through, then the fix is obvious,  Might also be worth reformatting the case a bit.
https://github.com/freebasic/fbc/blob/8271887e8cfd69722662234df7f292708aec6f13/src/gfxlib2/win32/gfx_win32.c#L329

countingpine:
I've submitted a couple more changes - some warnings appeared the next time I did a clean make, some from your list didn't appear at all, so I looked through and changed them all myself.  Hopefully I didn't miss any..

Looking at the win32 gfx code, I guess I can now say that WM_SIZE falls through to WM_SYSKEYDOWN (to handle their Maximize events), and WM_SYSKEYDOWN falls through to WM_KEYDOWN (to handle key events), but WM_SIZE will not fall through to WM_KEYDOWN.

It might be that it can be refactored - the `WM_SIZE` case could set the value of `is_maximize`, then break; the `WM_SYSKEYDOWN` case could set the value of `is_alt_enter`, then fall through; then `is_maximize || is_alt_enter` could be checked after the end of the switch block.  It would change the order of execution though.

Otherwise the simplest solution is to put a comment in saying something like `/* fall through for WM_SYSKEYDOWN /*`, which makes the flow clearer, and will probably satisfy GCC.  The final `if ... break` could also be pulled out of the scope.

countingpine:
Hopefully we've caught all the warnings now.  I am committing this.  Thanks both for taking the time to look at this.  It's been helpful.

PR #107, opened 2018-09-26
Allow suffixes in comments

jayrm:
- fixes sf.net bug #866 fbc throws lexer errors in comments stating with $


jayrm:
In the first commit,
there is probably a cleaner way to fix this, to allow suffixes in comments, but it gets the job done to fix sf.net bug 866

In the second commit,
I noticed that lexing of comments varies between dialects. In -lang fb, single line comments are treated as single line comments only. In the other dialects, single line comments are lexed for "/'" multi-line comments. I don't see any reason it should vary between dialects.  This change may break old codes having multiline comment comment tags "/'" starting in single line comments.

PR #110, opened 2018-10-22
C backend: struct size fixes (#858)

swurzinger:
Fixes #858: C backend: fix internal structure size mismatch due to wrong padding when nesting packed structures

The alignment of nested structures which are packed with a smaller alignment than the natural or specified alignment of the parent structure has to be specified explicitly, otherwise the field will be packed too.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -80,6 +80,7 @@ Version 1.06.0
 - #699: fix new[0] causing infinite loop when calling constructor/destructor list
 - #883: when mapping 32 & 64 bit functions for PALETTE [GET] USING, check the data type pointed to and choose one of LONG PTR, LONGINT PTR, INTEGER PTR
 - #866: fbc was throwing lexer errors in comments stating with $.  Comments are lexed for directives; allow suffixes in comments
+- #858: C backend: fix internal structure size mismatch due to wrong padding when nesting packed structures
 
 
 Version 1.05.0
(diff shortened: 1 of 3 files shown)


swurzinger:
I changed the test to work on systems with 4-byte QWord-alignment (like Linux) rather than 8-byte alignment.

dkl:
I tested the old test case (with long+longint) from SourceForge ticket #858, and it indeed causes the error with -gen gcc for linux-x86_64, win32 and win64 (not for linux-x86). But the new test case (with byte+short) seems to work fine.

We should probably keep both as regression tests. ABI differences between platforms can be handled with #if statements (e.g. different #asserts for each target) in the test case.

swurzinger:
I tried on win64 only and it seems that's the only platform where the modified tests work.

swurzinger:
I also tried on win32 and ubuntu-x86_64 and could reproduce the issue with the new test code as well. Are you sure that you cannot reproduce it on these platforms with the latest changes?

Maybe I should rather create a log-test instead of a unit-test?

dkl:
Hmm, my mistake, I tested with just the type declarations, without the Dim statement using the type, so the struct didn't get emitted by -gen gcc. After fixing that everything works (fails) as expected.

swurzinger:
The issue addressed by this PR, https://sourceforge.net/p/fbc/bugs/858/, can be closed, I think.

Issue #113, opened 2018-11-18
Bottom-right corner of Window Screen is transformed out of the viewport


countingpine:
See also: https://freebasic.net/forum/viewtopic.php?f=3&t=27168

In the below code, a white border should be drawn between two grey borders, but some of it is cropped out of the viewport, and so not drawn.
```
screen 12
view (120,120)-(220,220), , 7      '' grey border outside viewport

window screen (0,0)-(1,1)
line (0.01,0.01)-(0.99,0.99), 7, b '' grey inner border of Window (1px inwards)
line (0,0)-(1,1), 15, b            '' white outer border of Window (buggy line)

sleep
```
In QBASIC, all three borders can be fully seen.

I've investigated in https://freebasic.net/forum/viewtopic.php?p=254681#p254681, and found two bugs in `fb_hTranslateCoord()` (and associated code).

countingpine:
The above pull request #114 was accepted, meaning this has been fixed now.

PR #114, opened 2018-11-18
Fix two WINDOW scaling problems (GitHub Issue #113):

countingpine:
- coordinates should be multiplied by e.g. 'width - 1', not just 'width'.
- CINT() macro was not rounding positive values properly

Note: CINT() is also used in https://github.com/freebasic/fbc/blob/master/src/gfxlib2/gfx_circle.c#L96 and https://github.com/freebasic/fbc/blob/master/src/gfxlib2/gfx_draw.c#L247, so we need to make sure that they were not relying on the buggy rounding.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -83,6 +83,7 @@ Version 1.06.0
 - #858: C backend: fix internal structure size mismatch due to wrong padding when nesting packed structures
 - C backend: fix array descriptor mangling to avoid naming conflicts where array data types differ only by const
 - #823: Function overload resolution for [const] array() and passing non-const array argument to const array parameter
+- Github #113: When a WINDOW was active, the top/right edges (bottom/right for WINDOW SCREEN) were transformed out of the viewport.
 
 
 Version 1.05.0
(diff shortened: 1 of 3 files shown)


countingpine:
I think this change yields a positive effect in the rounding of DRAW, shown in this code:
```
screen 13
draw "bm160, 90 TA5" & string(50, "U")
draw "bm160,110 TA5" & string(50, "D")
draw "bm150,100 TA5" & string(50, "L")
draw "bm170,100 TA5" & string(50, "R")
sleep
```
Where previously there was a small "hook" at the start of the up/right spokes, this now no longer happens.  (This is because it now starts in the middle of the "step" instead of the start/end.)
Before:
unpatched
After:
patched

countingpine:
Similar improvements can be seen when drawing arcs with Circle.  Arcs are quite ugly anyway (no Bresenham), so I don't think the implementer would have cared much about the rounding method.
arc-unpatched
arc-patched
Note: in this case the patch only changes the rounding on positive x,y (the right/top sections of the circle).

I conclude that the patch is sound, and fixes less-than-ideal behaviour in other use cases, where the bug has not been accounted for.

dkl:
Even though I don't know the Draw code, the patch seemed logical. Especially the change to CINT(), which changes the casting (if I understood that correctly...). It makes sense to cast to int after the + 0.5, instead of before, for rounding.

PR #117, opened 2019-01-20
V 0.8.5.0 - (01-20-2019 13:24:10):  commit for pull request

jklwn:
Please read "ustring.txt" and "array.txt"

Short Summary:

A new data type for FreeBASIC: USTRING

USTRING is a zero terminated dynamic wide (unicode) string type, which can be used like any other available string type in FreeBASIC. All syntax elements working for strings, work for USTRINGs exactly the same way. USTRING is UTF16 Windows (not handling surrogate pairs) and UTF32 in Linux.

You must include "ustring.bi" into your code in order to use it. The underlying class-like type is an adaption of José Roca´s CWstr dynamic wide string type in WinFBX reduced to the absolute minimum. USTRINGs work for 32 and 64 bit in Windows and Linux.


jklwn:
My development system is Windows 7 Home. I can compile 32 and 64 bit versions and run all tests without errors. I have Ubuntu64 on a virtual machine and can compile and run the 64 bit version for Linux there. 

Currently i don´t know why Travis fails

JK

rversteegen:
Wow. Wasn't expecting to see something this. Has this been discussed on the forums? I haven't been watching them.
I was hoping someone was working on a built-in UTF8 type, but if that happens I guess it could be called U8STRING.

You can click on "Details" and read the log for the automated check. It fails because you forgot to check in array.bi, and also ustring.bi. (Also, you shouldn't have commented-out jk-ide specific changes (to fbc.bas) in there at all, and don't add comments to the code saying why you changed something ("'changed "name_" to "name_x" (avoid naming conflicts)") - that belongs in the commit message. Not to mention huge amounts of irrelevant whitespace changes mixed in.)

It appears that what you're doing is implementing USTRING as a UDT, and not really adding it as a built-in type. It's not for me to decide whether that's a good idea or not. But why is there a need for env.clopt.ustring?

jklwn:
Well, not all checks actually failed. Half of them passed the other half with "FBTRAVIS_COMPILER_DEBUG=1! keep failing.

The whitespace changes are irrelevant - yes. I didn´t intentionally mix them in and i don´t know how to avoid them. I´m fairly new to GIT and i´m struggling in general a bit with GIT not doing exactly what i want it to do. My development system is Windows.

I´m in contact with Jeff, there hasn´t been a discussion on the forum so far.

env.clopt.ustring is a switch for the added code. You must add "#pragma ustring" before you can make use of it. This can be used for debugging purposes: is a possible bug present with and without the new code, or does it show only, when new code is switched on?

JK

jklwn:
Finally everything seems to work as expected!

jklwn:
Well, the problem with GIT basically are the line endings. Cloning and checking out leaves files with LF line endings, which is a problem in Windows. So i had to normalize these files (e.g. emit.bi) some other way in order to be able to use them. I really tried a lot (btw. autocrlf=true doesn´t solve the problem) currently i think a .gitattributes file could be a solution, but i´m still trying to implement it in this fork.

I made this pull request to open a discussion, i don´t expect that it is accepted without any changes. I´m a grown up guy and can stand criticism - just say what´s wrong, no need to be overly polite.

The UDT/Class/Type (whichever way you like it) is based on code of José Roca, which is available in his WinFBX library for Windows. I know José from PowerBASIC and he agreed to my attempt for a better integration of his work into the compiler. Ustrings are implemented as CWstr in Paul Squires´ WinFBE editor for FreeBASIC for quite some time, so it can be taken for granted that the type itself basically works bug-free. There is one drawback José had to find a workaround for, the intrinsic string handling functions of the compiler cannot handle UDTs, because of the way the compiler has been coded. So his workaround is to prepend "**" to every ustring variable or expression, which in turn presents a dereferenced wstring pointer to the compiler - something it can deal with.

My primary intention is to get rid of the need for prepending "**". All changes to the compiler are related to this. In other words: adapt the intrinsic string handling functions in a way that an UDT, which can cast to wstring, is accepted and processed, just like a wstring. 

Ustrings are implemented by adding ustring.bi to your code, Jeff approved this approach as a first step. A full integration could follow with future releases, but for a start, this would do.

Stringex.bi and array.bi expand on ustring handling, but may be used with other string types (array.bi - all types) too. These functions proved to be of enormous help in PowerBASIC, so why not port them. There is a small adaption in the compiler in order to make the "ANY" keyword possible in some of the string handling functions just like in "INSTR". 

Memcpy and the default memory allocation and grow size can easily be adapted to what´s considered reasonable. It´s the usual tradeoff size vs. speed. The more you allocate beforehand the less (time consuming) copy operations you will need, when the length of the data changes.     

maybe a stupid question: not portable - not portable to what ? It works on Windows and linux both 32 and 64 bit.

JK

jklwn:
It´s so frustrating, everything i try doesn´t seem to make it better !

jklwn:
I will give up on it for the moment! Now GIT wants to delete all files in my repo, seems i managed to mess it up totally. Maybe i will delete my fork, rebuild it from scratch and make another pull request - i don´t know at the moment. Fortunately i didn´t rely on GIT, i got backups...

JK

rversteegen:
Line endings: this can be a huge pain; I remember them being a pain with svn too although svn handles them differently. autocrlf is trouble.
Without a doubt, the best solution is to leave all the line endings as LF, and just use a code editor that has no trouble editing them (and leaves them as they are), and don't set any of the git options for autoconverting line endings. Normally I would say "Only Notepad can't handle Unix line endings" but I'm not trying to insult anyone's IDE! So WinFBE can't handle them?

Your last commit changed all the line endings; you can get rid of it with `git reset --hard HEAD~1` followed by `git push -f`. NOTE: this will delete all changes you've made to your source files since then.

Not portable: only works on x86/64 DOS/Windows/Linux. FB supports ARM and probably any little-endian CPU (though I don't remember anyone using any other CPU). It won't work on Mac either, because the Mac port of FB uses at&t syntax assembly, not intel, because the Mac gas assembler doesn't/didn't support intel syntax properly. Thankfully Apple replaced that awful assembler with llvm.

Also, I notice the `PathName_` function is Windows-specific.

Speed/time tradeoff: pre-allocating a 1040 byte buffer isn't going to buy you speed if you're working with small strings, it's going 

String and array functions: I agree with you that FB needs useful functions like these. They're nice functions. It's absurd that FB only's builtin functions operating on arrays are REDIM, ERASE, LBOUND and UBOUND. Arrays aren't even a first-class datatype. I noticed that hack (from fxm) you had to do to get the pointer to an array descriptor. I got so sick of FB arrays that I wrote my own "vector" library in C for 1-dimensional arrays.

But new builtin string and array functions belong in the rtlib (written in C), rather than in a header. Quite a few functions would be far more efficient if written in C.

rversteegen:
Also, is it possible for you to rename this pull request?

jklwn:
I will retry with a new pull request

PR #118, opened 2019-01-23
USTRING V 0.8.5.0

jklwn:
Second try, different approach again, hopefully this works


jklwn:
Well, as long as there aren´t builtin functions like that, it´s better than having nothing. Because of the implicit conversions for string types other than ustrings and the way it is coded, it is not as fast as it could be, but it´s fast enough.

I thought PathName_  would work for Linux paths ("/") too.

jklwn:
Investigating the line ending problems further i found that some files from fbc master have a few CRLF instead of LF only. Downloading fbc master as a .zip file and then unpacking it reveals this. The following screenshot shows all files in src\compiler with such orphaned CRLF line endings. Taking a look at them in a HEX editor you can see that. Furthermore the number of changed lines after normalizing them by an external script corresponds to the number of occurrences of "illegal" CRLF line endings. This seems to be part of the problem why autocrlf doesn´t work. The last screenshot shows all changed files after cloning and normalizing the above mentioned files. Even with "filemode=false" i keep getting chmod changes and other changes, i´m not reponsible for. 

No luck - seems i cannot upload .png or .zip files here, i keep getting "Something went really wrong, and we can´t process that file. Try again"

jklwn:
Ok, i cannot upload here for whatever reason, but i can post a link (https://drive.google.com/drive/folders/1UkDEzigYJ3Qgtqr30JpcB68EGXUN0l1n?usp=sharing) to a folder of my google drive, where you can see the screenshots. "Changes zip" contains all files i changed or added code or text, you might just unzip it to your working directory, the folder structure fits.

rversteegen:
Good spotting. Someone should fix that.

The problem with pathname_ is that it takes \ and : to be path separators, but they aren't on Unix.

rversteegen:
And yes, it's better than nothing. FB needs a better standard library, it hasn't been expanded much since QB. So maybe some functions could appear first in a header and later, as people have time, get converted into being built into the rtlib and compiler?

jklwn:
A list of fixed files is under it´s way to Jeff.

So i must have two versions of PathName_, one for Windows and one for all others - or even more ?

Making these functions intrinsic ones would require changes to the compiler AND the runtime. Even if i had less problems than initially thought to apply the desired changes to the compiler, i will have to learn quite a lot about the compiler´s code, before i can do that. Let´s wait and see what the developers say about my current work.

JK

jayrm:
General comments:
JK, when we started to discuss these changes (mid Dec 2018) we identified the following, IIRC:
1) built-in dynamic WSTRING probably still a ways off (long time).  We would be duplicating much of the STRING code, and STRING still needs improvement.  We don't want to duplicate STRING problems with a new DWSTRING type, plus it's just a lot of work.
2) Implicit conversion to [W]STRING for user defined types (UDT's) when used with built-in string functions.  Currently, fbc prefers to implicitly convert UDT's to STRING if it has the OPERATOR CAST() AS STRING overload, but does not do the same if it has cast to wstring ptr (byref wstring)
3) Solving general case where implicit CAST results in ambiguous call or ambiguous function overload.  As can happen with UDT's & CAST.  Something along the lines of a DEFAULT CAST overload.

Specific:
- So, I got about as far as 2) Implicit conversion to [W]STRING for user defined types (UDT's) when used with built-in string functions.
- This pull requests adds much more than just that.  And I haven't had time to review the other additions in any kind of detail.  It's going to take some time to review everything.  Though, if we post your changes (like here), or make binaries available to users, you might be able to get more feedback than just a couple of developers.
- The CRLF issue is separate, and has to do with how files are stored in the git repository.  We do want to fix it  That said, text editors and IDE's should be accomodating about what files they can read:.  My MS DevStudio 97 can read LF files just fine and only modifies the file to CRLF if I actually make edits to the file.
- These are all new to me: "TALLY", "PARSECOUNT", "PARSE_", "EXTRACT_", "REMAIN_", "REMOVE_", "REPLACE_", haven't had a chance to look at much.  Look like features that are available in PB.  This might all be just fine, but so far, you are probably the only one that has tried it out.

Bottom line:
Going to take some time to go through this pull request fulll.  The first priority (what I have been busy with) is the next 1.06 release, which i basically based on the code from Nov 2018.  We want a nice stable release right now.  After the release is done, then maybe can go through this and get it merged in, or divide it in to small chunks, which may be more manageable to review and commit.

jklwn:
Stringex.bi and array.bi are extensions for FreeBASIC implementing (the new type) ustring (you could see it as additional test piece too). The more, all string handling functions are written to work with zstring, string, wstring and ustring as well. Due to implicit conversions this is not the fastest way, but the mentioned overloading problems (ambiguous ...) make code specific for each type impossible for the moment. These functions are available in PB and adapted from José´s (Josep Roca) WinFBX (which is for Windows only), Paul (Squires) makes use of them in his WinFBE Editor for quite some time. So basically we know these functions are working (if i didn´t manage to introduce bugs when adapting them, which aren´t revealed by the added tests). José has more of those functions, but for a start i included, what i considered most useful. With a lot of effort these functions could be made (even faster working) intrinsic functions.  

Array.bi works with all types of variables and has (except for ustring) no dependecies to compiler changes.

Stringex.bi required a small change in the compiler´s code in order to make the "ANY" key word for some of the functions work just like with "INSTR". All other changes in the compiler address #2 of our list where ever needed. Hopefully it covers all places.


There are no changes to overload resolving so far (#3). I second a step by step approach, therefore the amount of current changes seems to be enough to test and possibly merge for a first step. This step is about bringing a dynamic wide string type to FreeBASIC (without the need of already existing workarounds like "**") at all.

The overload problem isn´t specific to strings in general, you can get this annoying (ambiguous...) with other types too. Being able to use an UDT, which can cast to string and to wstring, would be no gain for the current changes, because implicit conversion works perfectly well. Having to do that by code inside an UDT (convert string to wstring and vice versa) wouldn´t be any faster. So we can make this (getting rid of "ambiguous" by introducing a default for overloaded functions or casts) a next step.


Reading files with LF line endings isn´t a problem, but editing is, because then normalization to CRLF is taking place in Windows as you say yourself. But you got my proposals for fixing this Windows/Linux problem in the repository satisfying and for good.

>>>Though, if we post your changes (like here), or make binaries available to users, you might be able to get more feedback than just a couple of developers.

Why not! It´s three new include files (ustring.bi, stringex.bi and array.bi) and the different compilers for 32 and 64 bit Windows, Linux, etc. plus documentation. How should we do that? I think i cannot attach executables to posts in the forum?

JK

jayrm:
OK, thanks, let's follow this up after 1.06.0 release.  I have reviewed some of the changes.  There are some changes which I support, and some which I think need further clarification & discussion.  Thanks.

jayrm:
I think this pull request incorporates too many changes in a single pull request:
- The Array features should be a separate pull request
- The StringEx features should be a separate pull request
- the UString.bi implementation should be a separate pull request

For the "USTRING" changes I recommend closing this pull request in favour of 
PR #150 which:
- uses existing syntax `EXTENDS` instead of `#PRAGMA`
- has multiple commits to track development progress
- has many more tests written
- focuses on extending UDT's to work as a `zstring` or `wstring`
- focuses on compiler features/changes rather than implementation (i.e. `ustring.bi`).

Forum discussion at [Extending Wstring and Zstring with UDTs](https://www.freebasic.net/forum/viewtopic.php?f=17&t=27569)

jklwn:
Ok Jeff, i agree, i will close this pull request in favour of yours (#150) and i will open new requests for Arrays, StringEx and UString.bi in the future.

PR #125, opened 2019-03-12
Fixed bugs

skyfish4tb:
Fixed bug: If Char with wchar connection, char contains double byte characters (such as Chinese characters), the "wstr = char & wchar" will becomes "wstr = char & Chrw (0) & wchar".


jayrm:
Thanks @skyfish4tb.

If I understand correctly, the double byte char from source buffer to wchar conversion should always have a length (in characters) that is equal to or less than original length (in characters) so should never overflow the allocated destination buffer.  This looks OK to me.

The "Merge pull request # 33" from has no meaning in this repository.  It's not needed for this commit.  I cherry picked the commit, and added commit message commit on to fbc/master.
This fix added in commits: f008be8257a50171ed37f085f3803b132c8dc404 and 5272e3b07617e8cdd2d40a597547c0b072bc6d49

PR #126, opened 2019-03-12
Fixed mismatched matrix mode of matrix stack push/pop calls

gothon:
Since each OpenGL matrix has is own stack, we need to ensure the correct stack is active before popping its stack.  I added calls to handle both GL_PROJECTION and GL_MODELVIEW matrix stacks in addition to ensuring that the MatrixMode() of the PopMatrix() call match the PushMatrix() call.

### DIFF
--- a/src/gfxlib2/gfx_opengl.c
+++ b/src/gfxlib2/gfx_opengl.c
@@ -266,6 +266,7 @@ void fb_hGL_SetupProjection(void)
 
 	__fb_gl.Ortho(-1, 1, -1, 1, -1, 1);
 	__fb_gl.MatrixMode(GL_MODELVIEW);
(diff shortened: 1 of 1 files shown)


jayrm:
This looks OK to me.  Thanks @gothon.

FYI, I wanted to add a message in `changelog.txt` so I tried a different technique for merging the patch and closing the pull request to see how it would work out:
`git fetch origin pull/126/head:gothon_pr126`
`git checkout gothon_pr126`
I then committed changelog.txt, with a changelog message and finally pushed to fbc:,master.
`git push origin gothon_pr126:master`
It had the desired effect, sort-of.  The pull request is merged and automatically closed, however, did not have the usually merge commit (message) that is generated when using the github interface.

Issue #131, opened 2019-04-05
"make bootstrap" race condition


vilhelmgray:
The following error occurs when building from the [FreeBasic-1.06.0-source-bootstrap.tar.xz](https://github.com/freebasic/fbc/releases/download/1.06.0/FreeBASIC-1.06.0-source-bootstrap.tar.xz) source code:

    LINK bootstrap/fbc
    AR lib/freebasic/linux-x86_64/libfb.a
    gcc: error: lib/freebasic/linux-x86_64/libfb.a: No such file or directory
    AR lib/freebasic/linux-x86_64/libfbpic.a
    make: *** [makefile:1093: bootstrap/fbc] Error 1
    make: *** Waiting for unfinished jobs....
    AR lib/freebasic/linux-x86_64/libfbmt.a

The following command was used to run the build:

    make -j5 bootstrap CFLAGS="-DDISABLE_X11 -DDISABLE_GPM -DDISABLE_FFI -DDISABLE_OPENGL"

I suspect this is a race condition issue, because the issue disappears when the number of make jobs is restricted to 1:

    make -j1 bootstrap CFLAGS="-DDISABLE_X11 -DDISABLE_GPM -DDISABLE_FFI -DDISABLE_OPENGL"

On my system, this issue is only present when making bootstrap; making all (i.e. compiling fbc normally using an existing fbc) completes correctly even with multiple make jobs.

jayrm:
Fixed in pull request #132.  Thanks.

PR #132, opened 2019-04-06
makefile: Fix bootstrap recipe race condition

vilhelmgray:
Since fbrt0.o and libfb.a are statically linked to the bootstrap fbc,
rtlib should be listed as a dependency of BOOTSTRAP_FBC. This patch
fixes the race condition described in issue #131.

### DIFF
--- a/makefile
+++ b/makefile
@@ -1063,7 +1063,7 @@ bootstrap-dist:
 #
 BOOTSTRAP_FBC := bootstrap/fbc$(EXEEXT)
 .PHONY: bootstrap
-bootstrap: rtlib gfxlib2 $(BOOTSTRAP_FBC)
+bootstrap: gfxlib2 $(BOOTSTRAP_FBC)
 	mkdir -p bin
 	cp $(BOOTSTRAP_FBC) $(FBC_EXE)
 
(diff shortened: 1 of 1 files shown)


countingpine:
Yeah, I think this makes sense.
Probably better for @jayrm to give the official verdict, so I'll give him a chance to weigh in, but it looks good to me.

vilhelmgray:
I came across this issue while compiling a bootstrap FreeBASIC compiler to compile the FreeBASIC compiler in [1.06.0.tar.gz](https://github.com/freebasic/fbc/archive/1.06.0.tar.gz):

    cc => bootstrap_fbc => fbc

I noticed that for this use case, compiling _gfxlib2_ is unnecessary because the _fbc_ source code does not depend on it. Would it make sense to remove the **gfxlib2** dependency from the **bootstrap** make recipe in this patch?

Would it instead be better to open a second pull request to implement a new make recipe for this scenario: **bootstrap-minimal** which compiles a bootstrap compiler with the minimal features (e.g. _rtlib_ dependency) necessary to compile another _fbc_?

jayrm:
@vilhelmgray, yes, makes sense to me also.  Thanks for finding this.  I will merge in.  Since this pull request specifically fixes issue #131, making a separate pull request for **bootstrap-minimal** sounds like a good approach to tracking the changes.

The only time fbgfx library is needed is if any gfx function used, or user explicitly includes fbgfx.bi.  The fbc compiler does neither, so a **bootstrap-minimal** would be a nice additional option for the builder.

Issue #137, opened 2019-04-21
Final -Wl flag takes precedence over previous


vilhelmgray:
When invoking **fbc** with multiple `-Wl` flags, only the final `-Wl` flag is passed down to the linker.

To demonstrate the issue, I have run **fbc** in 4 different ways:
 * no `-Wl` flags
 * one `-Wl` flag with two **ld** options (one defines symbol _foo_, other defines symbol _bar_)
 * two `-Wl` flags each with one **ld** option (_foo_ symbol definition first)
 * two `-Wl` flags each with one **ld** option (_bar_ symbol definition first)

After each compilation I run **nm** on the resulting executable and **grep** for either of the _foo_ or _bar_ symbols:

```
$ fbc -g foobar.bas && nm foobar | grep "foo\|bar"
$ fbc -g -Wl --defsym=foo=0,--defsym=bar=0 foobar.bas && nm foobar | grep "foo\|bar"
0000000000000000 A bar
0000000000000000 A foo
$ fbc -g -Wl --defsym=foo=0 -Wl --defsym=bar=0 foobar.bas && nm foobar | grep "foo\|bar"
0000000000000000 A bar
$ fbc -g -Wl --defsym=bar=0 -Wl --defsym=foo=0 foobar.bas && nm foobar | grep "foo\|bar"
0000000000000000 A foo
```

As can be seen from the tests, only the last `-Wl` flag is sent down to the linker program. The expected behavior is to pass down all linker options specified by all `-Wl` flags, and not simply the final one; this would match the [behavior described by **ld** documentation](https://linux.die.net/man/1/ld):

> Note---if the linker is being invoked indirectly, via a compiler driver (e.g. gcc) then all the linker command line options should be prefixed by -Wl, (or whatever is appropriate for the particular compiler driver) like this:
> 
> gcc -Wl,--start-group foo.o bar.o -Wl,--end-group
> 
> This is important, because otherwise the compiler driver program may silently drop the linker options, resulting in a bad link.

jayrm:
Fixed in #138, Thanks.

PR #138, opened 2019-04-21
Pass down all options from all -Wa, -Wc, and -Wl flags

vilhelmgray:
All options from all -Wa, -Wc, and -Wl flags are passed down to their
respective programs. This fixes issue #137.

### DIFF
--- a/src/compiler/fbc.bas
+++ b/src/compiler/fbc.bas
@@ -1869,13 +1869,13 @@ private sub handleOpt(byval optid as integer, byref arg as string)
 		end if
 
 	case OPT_WA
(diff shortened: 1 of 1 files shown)


jayrm:
Nice.  Makes perfect sense.

fbc documentation doesn't say one way or another what it will do, so will add a note there.
[Compiler Option -Wa](https://www.freebasic.net/wiki/wikka.php?wakka=CompilerOptWa)
[Compiler Option -Wc](https://www.freebasic.net/wiki/wikka.php?wakka=CompilerOptWc)
[Compiler Option -Wl](https://www.freebasic.net/wiki/wikka.php?wakka=CompilerOptWl)

Update to changelog.txt will follow in a separate commit.

Thanks

Issue #130, opened 2019-03-23
FreeBSD - fre bug


fatman2021:
fre returns 0 in FreeBSD 12.0-RELEASE.

rversteegen:
The implementation of FRE in FreeBSD is here: https://github.com/freebasic/fbc/blob/master/src/rtlib/freebsd/sys_fmem.c

As a FreeBSD user, perhaps you can work out what's wrong with it?

vilhelmgray:
I'm willing to test this to verify if `FRE` still returns 0 in FreeBSD 12, but I'm not sure how to bootstrap a FreeBSD version of **fbc** -- it looks like the [FreeBASIC bootstrap tarball](https://github.com/freebasic/fbc/releases/download/1.07.0/FreeBASIC-1.07.0-source-bootstrap.tar.xz) is missing the FreeBSD code under the `bootstrap/` directory. Is bootstrapping still supported for FreeBSD?

dkl:
No, but I would expect doing what [make bootstrap-dist](https://github.com/freebasic/fbc/blob/master/makefile#L1051) does should work for going from Linux to FreeBSD too. (see also https://freebasic.net/wiki/wikka.php?wakka=DevBootstrap)

rversteegen:
Bootstrapping works fine for FreeBSD, but you need to edit the makefile. In fact it works better than compiling from FreeBSD, which doesn't work (at least not with -gen gcc, haven't tried on x86 FreeBSD)

See https://github.com/rversteegen/fbc/commits/freebsd
Sorry, I need to make a pull request for those fixes. I'll try in coming days.
I do have FreeBSD 11 in a VM and can have a look at this bug too.

vilhelmgray:
I compiled the following FreeBASIC program in x86-64 FreeBSD 12 using FreeBASIC 1.7.0:
```
Dim mem As Integer = Fre

Print "Free memory:"
Print
Print mem; " bytes"
```

After running the program, it successfully printed out my free memory. I was unable to recreate the issue described in the original report.

@fatman2021 Are you able to confirm that this issue is still present for you using the latest FreeBASIC release?

jayrm:
@fatman2021,
Same here, I was unable to recreate the issue described in the original report.

Otherwise, `fre` returns something that seems reasonable (compared to value returned by `$ sysctl hw.usermem`), in fbc 1.07.1 built for:
- FreeBSD 12.0-i386
- FreeBSD 12.0-amd64
[...]

PR #139, opened 2019-04-23
rtlib: freebsd: Fix deprecated use of VM_METER

vilhelmgray:
VM_METER has been deprecated for over 15 years; VM_TOTAL should be used
to request the population of a struct vmtotal.

### DIFF
--- a/src/rtlib/freebsd/sys_fmem.c
+++ b/src/rtlib/freebsd/sys_fmem.c
@@ -7,7 +7,7 @@
 
 FBCALL size_t fb_GetMemAvail( int mode )
 {
-	int mib[2] = { CTL_VM, VM_METER };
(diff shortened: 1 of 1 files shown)


rversteegen:
Is this a fix for gh bug #130 ?

vilhelmgray:
> Is this a fix for gh bug #130 ?

I suspect #130 is a different issue. Despite the deprecation, `VM_METER` is still mapped to `VM_TOTAL` in the current FreeBSD source code, so this change should not have an effect on the current behavior of the FreeBASIC FRE command: https://github.com/freebsd/freebsd/blob/master/sys/vm/vm_param.h#L78

jayrm:
Thanks, update to changelog.txt will follow in a separate commit.

Issue #140, opened 2019-04-24
Implement control over symbol stripping


vilhelmgray:
I'm developing an **fbc** package for Gentoo, but there's a behavior in **fbc** that causes problems for the Gentoo build system. Gentoo packages are required to build executable with the symbol table intact (i.e. "_not stripped_" executables).

However, **fbc** automatically passes the `-s` option down to the linker if debugging/profiling is not enabled; this strips all symbol information from the executable. Since the addition of the `-s` option cannot be controlled, **fbc** executables cause the Gentoo build system to fail.

It is true that symbols will not be stripped if debugging is enabled, but this bloats the executables with additional information that is not used (e.g. *debug_info*). There are situations where symbols are useful outside of source code debugging, where *debug_info* and such are not neeeded.; for example, a live memory analyzer can keep track of symbol values if they are simply present in the symbol table without the need of additional debugging info.

I propose the following new compiler option:

```
  -strip           Omit all symbol information from the output file
```

It will be useful to have a way to manually control whether symbols are stripped or not. That should give user much finer control and freedom over the resulting executables.

dkl:
That reminds me of https://www.freebasic.net/forum/viewtopic.php?f=17&t=22958 which was about getting a "release" build but with debug info and symbols, which seems to be somewhat similar.

Anyways, an option to disable the -s during linking sounds useful for that. It should probably remain the default for non-debug builds though.

jayrm:
@vilhelmgray, is the change desired for just building the fbc package?  Or is change intended to be global, like for all fbc built executables?  If we add '-strip' option only, what is the justification that not stripping symbols by default is the way to go on all platforms for all users?

Most users typically don't seem to care about executable symbols or the benefits and are only looking at binary file size, so stripping all symbols was selected as default.

If it's only desired for building fbc package, then can achieve same result with a '-nostrip' option, and apply through fbc's makefile.

If preserving symbols is the norm on gentoo linux and meant for all executables, ones built by fbc also, AND we want to keep the strip all symbols default on other systems, then it's more like a configuration option for the fbc package build selected by the package maintainer.  In which case, we probably need with both a '-nostrip' and a '-strip' option available, where one is selected by default depending on the package build options.

vilhelmgray:
That's a good question, whether all executables compiled by fbc by default should be stripped or not. I decided to investigate why Gentoo requires symbols, and why **gcc** does not strip all symbols by default.

[...]

PR #141, opened 2019-04-24
Implement the '-strip' compiler option

vilhelmgray:
This change allows users to choose whether or not to strip symbol
information from the output file. The fbc '-strip' option is analogous
to the ld '--strip-all' option.

This fixes issue #140.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,9 @@ Version 1.07.0
 [added]
 - CVA_LIST type, CVA_START(), CVA_COPY() CVA_END(), CVA_ARG() macros will map to gcc's __builtin_va_list and __builtin_va_* macros in gcc backend
 - github #133: fbc makefile supports bootstrap-minimal target to build a bootstrap fbc with only the minimal features necessary to build another fbc. (William Breathitt Gray)
+- github #141: introduced '-strip'/'-nostrip' options to control symbol stripping of output files (William Breathitt Gray)
+- github #141: fbc will default to stripping symbols if '-d ENABLE_STRIPALL' is passed in FBCFLAGS (William Breathitt Gray)
+- github #141: makefile option 'ENABLE_STRIPALL=1' introduced to pass '-d ENABLE_STRIPALL' via FBCFLAGS by default for dos/win32 targets (William Breathitt Gray)
 
 [fixed]
 - sf.net #881: C backend: support for varadic function parameters in gcc using __builtin_va_list type and related macros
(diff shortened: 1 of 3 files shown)


vilhelmgray:
Would you like me to add changelog.txt changes as part of this commit, or separate?

dkl:
Ideally in the same commit.

jayrm:
I agree.  Ideally, updates to changelog.txt:
- in the same commit if the pull request is a single commit
- in the most recent commit (can be separate) if the pull request is multiple commits.
It is helpful to see updates to changelog.txt as part of the merge commit when merging the PR to master.

----
Not related specifically to this PR: a few of my thoughts follow, as I've been trying to take notice how we and github handle certain scenarios.

Whenever we have multiple pull requests on the go, we almost always get conflicts in changelog.txt.  I have tried github's method to resolve conflicts on-the-fly: it works, but it does a merge master to PR, and triggers another Travis-CI build.  Not the best in my opinion as it adds the extra merge commit just to resolve changelog.txt.

Rebase before merge seems like it should be OK in my opinion, resolving the conflicts locally.  However, obviously must force push the new PR, and I haven't discovered the benefit of github keeping track of the force pushes yet.  Also triggers another Travis-CI build.

A strategy for handling changelog.txt without conflict would be helpful.  But I don't know what that strategy would be.

Regardless, `git blame changelog.txt` is a useful search method no matter where the changelog.txt gets updated, even directly on master branch, provided that updates are in same commit or immediately after the relevant commits made.

vilhelmgray:
I've forced-pushed an updated commit with the changelog.txt changes.

Regarding your Github pull request workflow, it would be best to stick with the "Rebase and merge" option as it will keep the git history of the master branch clean without all the superfluous merge commit entries. This will benefit contributors looking back on the development history of the project, and ease debugging via `git bisect` operations and such.

The canonical way to achieve this in a pure git setup is to rebase on the local branch before fast-forwarding the master branch. In Github, this will create a new force-push entry, which triggers another Travis-CI build. Since the force-pushed commit after rebase would in theory be new -- the prescriptive context in the diff changed to accommodate the conflicting changelogs.txt  of the other PR -- then another Travis-CI build would be warranted on the admittedly rare chance that some code was clobbered during the merge conflict resolution.

The reason force-pushes are tracked by Github is an extension of the Github site-wide philosophy of "always retain development history". It's the same reason why pull requests and issues are closed and not deleted. It's somewhat of a nuances to track obsolete force-pushes, but I think we're stuck with it unless Github changes its policy. Though it may be useful to have retained the force-push history if we end up with a situation where the merge conflict resolution introduces a bug by accident.

jayrm:
Thanks.  Fixes #140

Also, thanks for the helpful feedback on our workflow.

PR #150, opened 2019-06-02
user defined types can extend zstring or wstring

jayrm:
Allow a user defined type (UDT) to inherit properties and behaviours of a zstring or wstring by extending existing syntax.  Purpose is to allow users to create custom string types (i.e. dynamic memory management) that can integrate well in to existing fbc compiler built ins. 

New Syntax:
```
type UDT extends zstring|wstring [, base-type]
end type
```

In places where fbc compiler would have normally rejected a UDT, this declaration of a UDT will instruct compiler to convert the UDT to a `z|wstring` through a suitable `CAST` operator.

`zstring|wstring` behaviour can be inherited directly from `zstring|wstring`, or indirectly and singly from the base type:
```
type B extends wstring
	'' member data
	'' inherits from wstring
end type

type D extends B
	'' D inherits from B and wstring
end type
```
Or `zstring|wstring` behaviours can also be inherited in a UDT with a kind of pseudo multiple-inheritance:
```
type B
	'' member data
end type

type D extends wstring, B
	'' D inherits from B and wstring
end type
```

Changed:
- SADD/STRPTR(wstring) returns WSTRING PTR

Added:
- 'TYPE udt EXTENDS Z|WSTRING' allowed to specify that UDT is a kind of Z|WSTRING
- LTRIM/RTRIM/TRIM will accept UDT as Z|WSTRING
- LCASE/UCASE will accept UDT as Z|WSTRING
- Cxxx() conversion functions will accept UDT as Z|WSTRING
- INSTR/INSTRREV will accept UDT as Z|WSTRING
- MID function will accept UDT as Z|WSTRING
- SADD/STRPTR will accept UDT as Z|WSTRING to return Z|WSTRING ptr
- LSET/RSET statements will accept UDT as Z|WSTRING
- MID statement will accept UDT as Z|WSTRING
- ASC function will accept UDT as Z|WSTRING
- STR/WSTR function will accept UDT as Z|WSTRING to return a Z|WSTRING

The only change to existing behaviour is `SADD/STRPTR(wstring)` now returns `WSTRING PTR`.  This will break any user code where `STRPTR(wstring)` is expected to return a `ZSTRING PTR`

New implicit conversions and behaviours are activated only if a UDT extends `zstring` or `wstring`.

In future expect separate pull requests for:
- `LEFT`, `RIGHT` to accept UDT extends `z|wstring` without having to explicitly overload `LEFT`, `RIGHT`.  This involves fixing, in my opinion multiple string related bugs on sf.net
- `IIF` to allow handling UDT's that inherit `zstring|wstring`
- `SELECT CASE` to allow handling UDT's that inherit `zstring|wstring`
- `SWAP` which maybe would be handled best with an overloaded member function rather than implicit conversion to `zstring|wstring`

This pull request introduces the initial feature and handles nearly all the quirk keywords.  I'd like to merge this change in before working on the remaining features in separate pull requests.

Forum discussion at [Extending Wstring and Zstring with UDTs](https://www.freebasic.net/forum/viewtopic.php?f=17&t=27569)

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -19,6 +19,7 @@ Version 1.07.0
 - LSET/RSET statements will accept UDT as Z|WSTRING
 - MID statement will accept UDT as Z|WSTRING
 - ASC function will accept UDT as Z|WSTRING
+- STR/WSTR function will accept UDT as Z|WSTRING to return a Z|WSTRING
 
 [fixed]
 - sf.net #881: C backend: support for varadic function parameters in gcc using __builtin_va_list type and related macros
--- a/changelog.txt
+++ b/changelog.txt
@@ -21,6 +21,7 @@ Version 1.07.0
 - ASC function will accept UDT as Z|WSTRING
 - STR/WSTR function will accept UDT as Z|WSTRING to return a Z|WSTRING
 - SELECT statement will accept UDT as Z|WSTRING to return a Z|WSTRING
+- SWAP statement will accept UDT as Z|WSTRING
 
 [fixed]
 - sf.net #881: C backend: support for varadic function parameters in gcc using __builtin_va_list type and related macros
--- a/changelog.txt
+++ b/changelog.txt
@@ -12,6 +12,7 @@ Version 1.07.0
 - LTRIM/RTRIM/TRIM will accept UDT as Z|WSTRING
 - LCASE/UCASE will accept UDT as Z|WSTRING
 - Cxxx() conversion functions will accept UDT as Z|WSTRING
+- INSTR/INSTRREV will accept UDT as Z|WSTRING
 
 [fixed]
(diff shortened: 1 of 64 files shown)


jayrm:
Force push to rebase on current fbc/master

jayrm:
This pull request now also includes additions for `SWAP`, `IIF`, and `SELECT`. to accept a UDT extends Z|WSTRING.

Existing user code is not affected by this pull request, and I plan to continue support for the feature in future.  On this basis I believe it acceptable to merge in to master without a review.

dkl:
Sounds good.
Since you're currently pretty much the only person working on the project, reviews don't make much sense (no point waiting for reviews if nobody is doing it). I try to stay up-to-date here or there, but then again, for me it was usually "whoever makes the change gets to decide" and so I've been trying to not comment too much.

PR #153, opened 2019-06-17
Ustring

jklwn:

In Windows you should implement WinFBX as a recommended alternative, but "ustring.bi" must be included (after the "AFX\..." includes !) nevertheless for proper working. "ustring.bi" contains code for 
activating the USTRING feature in the compiler and macros for switching the "USTRING" keyword to José´s "AFX.CBstr".

The more there are new definitions, which might be useful when dealing with strings:

"DSTR" dynamic string type,    = USTRING for UNICODE defined, STRING otherwise
"ZSTR" zero terminated string, = WSTRING for UNICODE defined, ZSTRING otherwise

This allows for writing code, which uses unicode string types for UNICODE defined and ascii string types otherwise the code will compile successfully in both cases.

All other changes are white-space changes converting leading spaces in a line to tabs (one tab replaces 4 spaces). This is an attempt to improve formatting consistency in the repository. Jeff, if you don´t like this, just ignore white-space changes.

JK


dkl:
The white-space changes should definitely be separate (if at all), since they are not related to this topic. In general we've been trying to avoid global whitespace changes to avoid "breaking" `git blame`.

jklwn:
Well, as already said above, this is an attempt to improve formatting consistency in the repository. In the beginning i had serious problems with line endings and thought it is my fault, because i didn´t have much experience with GIT. Then i realized, that in fact it was due to inconsistent line endings in the compiler´s sources. I spotted the offending files, fixed the line endings and sent them to Jeff. He obviously fixed it in the repository, because now these problems are gone. 

We had a discussion elsewhere about formatting rules for the sources. One of it was, indentation should be tabs only, one tab being 4 spaces. I keep coming across places, where indentation is not correct at all, or more often where it´s spaces instead of tabs. Honestly i tend to mix it up too ...

So i wrote a little utility, which converts leading spaces (for indentation) to tabs. What you see is the result. It´s not just for fun, it´s an attempt to make it consistent again, where it currently isn´t.

JK

Mr-Swiss:
Sorry, but I disagree 100%.
TAB's in source-files should be disallowed entirely.

The reason is simple: everybody's tastes differ and therefore, TAB's are completely unsuitable when a consistent environment is at a top priority.

Before any source-file is saved to storage (insert: Disk/SSD/Thumb-Drive, whatever), TAB's should forcefully be expanded, to the agreed uppon amount of SPACES (4 in this case).

jklwn:
I don´t want to argue, but that´s what Jeff stated here (http://www.jose.it-berater.org/smfforum/index.php?topic=5253.msg23799#msg23799). 

>>>In future, I would do the white-space changing commits separate from the content changing commits.

Ok, this is not exactly what he wanted, but there are no code changes at all except the mentioned spaces to tab conversions. I added "ustring.bi" and some tests in tests/ustring and i updated the file "dirlist.mk" for the added ustring directory. So it´s easy to keep things apart, e.g. take the ustring part and discard the rest, or accept the white-space changes as a measure to sanitize the code base. If you want to the latter, i know it will be a lengthy and boring task to re-check every change - just to be sure. But once done, you will have a cleaner code base. It´s up to Jeff ...

JK

rversteegen:
> The white-space changes should definitely be separate (if at all), since they are not related to this topic. In general we've been trying to avoid global whitespace changes to avoid "breaking" `git blame`.

You can pass `-w` to `git blame` to ignore changes to whitespace. (You're probably aware.) It is annoying though, especially when using other tools like git frontends and having to figure out how to pass that argument through to git.
But IMO, the inconsistent use of whitespace in the codebase is even more annoying. I'm used to reading the FB source with the indentation being totally broken because I can't be bothered changing my editor's tab width every time I open a fb/src/rtlib/*.c file.

Re: Mr-Swiss: due to that problem of varying tab widths, personally I use tabs only for indentation to the start of the line, and spaces elsewhere, eg to make line continuations line up. It's more work to do so though, so understand where your vitriol comes from.

Obviously the whitespace changes need to be split out to a separate commit. @jklwn: create a new temporary branch from upstream master, apply your whitespace cleanup to it, commit it, and then rebase this branch onto your temp branch (which you can then delete). That way all the whitespace changes will disappear from this commit (and jeff can drop the whitespace commit if he wants).

Mr-Swiss:
RE: rversteegen: "It's more work to do so though, so understand where your vitriol comes from."

I'm aware of that, but a decent IDE (like: FBEdit) can be configured, to automatically convert (aka: expand) TAB's in source, to SPACES uppon saving the file.

Btw. I'm using a similar approach, as described by you, when writing FB-code.
My reason to do so was mainly motivated by the FB-forum SW, which expands TAB's to a indentation of 3 SPACE's and always made my sources (4 indentation SPACES) look very messy. Since changing to expanding TAB's the problem has vanished.

jklwn:
Personally i dislike tabs too and i avoid them, where ever i can, but i´m not the one to decide this here.

You can use tabs or spaces for indentation and alignment. As long as you stay consistent (tabs OR spaces and if you use a fixed font), everything is fine. The fun begins, when you mix both and the tab width of your editor differs from the original setting. If you prefer a fixed font for coding, you could fix that by setting the tab width to the correct value. If you prefer a proportional font, alignment will never work with spaces, you must use tabs only. 

So my personal choice is, i always use a fixed font and i never use tabs (if possible). This way my code is nice to read, everywhere and regardless of the tab setting of the editor. 

This may not be the most urgent thing on our agenda, but it would be really nice to have a strict rule for this, and then make the code base compliant to this rule, even if this rule doesn´t meet my personal preferences (i would prefer spaces too). But Jeff wants tabs, so he gets tabs.

I actually expected a discussion about USTRING and not so much about tabs or spaces ...

JK

rversteegen:
USTRING is a nice addition, because the lack of variable length widechar strings in FB is a crippling shortcoming. If it's only UCS2 and not UTF16 on Windows though that is a problem but it can probably be fixed later. I need to have another look at the code, because I've forgotten anything about that. However I've already decided on and started using UTF8 instead of widechars.
Right now perhaps the biggest problem whether using WSTRING or UTF8 is that FB's pathname functions (e.g. DIR, OPEN) only accept 8-bit strings, which means the current ANSI codepage on Windows. It's less of a problem on Unix, where pathnames are always mbcs. Note that as an exception Windows pathnames are NOT UTF16, they are UCS2, and filenames with mismatched surrogate pairs do happen. Other programming languages like Python have had to jump through loops to translate these filenames to/from UTF8/UTF16/UTF32.

I think you two misunderstood me: I use mixed tabs and spaces to ensure the indentation is always consistent without needing to use spaces for indentation.
Never heard of anyone using a proportional font for code, except for comments. That's bizarre.

Anyway, point is, don't mix huge unrelated changes into pull request, or worse, into a commit, worst of all if it's flamewar fuel.

jklwn:
If you use tabs tabs in one line and spaces in the next for indentation and/or alignment, the problem is that in this case for a correct display you need a correct tab width setting. This stetting may differ with code or device (it´s not only the editor, which had to and could be adapted, it may be forum software and other things too, which cannot be adapted). In other words, even if you want you cannot get a correct tab width setting sometimes and all the work you put into your code for an easy to read and nice looking result was in vain. So my conclusion is, i use two spaces for indent (not too much type work) and i use spaces for alignment too, i never use tabs. This way i always and everywhere get a correct display regardless of the tab setting - i absolutely agree to Mr Swiss spaces on this point.

Wide strings are needed for Windows, because it´s API doesn´t natively support UTF-8, you can do conversions though. Currently the FB RTL converts wide strings to ANSI, before passing it to the C runtime. That is, you can have Ansi file names passed as wide strings, but you cannot have a real Unicode file name (e.g Russian characters). I have a working test version based on USTRING fixing this for OPEN and all other file or path name dependent statements. This will be a next step.

UCS2 covers most part of Unicode, UTF-16 (supporting surrogate pairs) is necessary for Asian languages like Korean, Japanese or Chinese, so for a start it would do.  

Sorry for mixing USTRING and white space changes. As said above it nevertheless could easily be kept apart. Maybe i´m too impatient and want too much in one time.

JK

rversteegen:
> I have a working test version based on USTRING fixing this for OPEN and all other file or path name dependent statements. This will be a next step.

Looking forward to seeing this! But instead of USTRING, don't you mean WSTRING, since USTRING is not native, so the rtl knows nothing about it?

jklwn:
Of course it´s WSTRING but you can pass USTRINGs too. This requires some changes in the compiler´s sources and mostly additions to the RTL.

JK

jayrm:
> The white-space changes should definitely be separate (if at all), since they are not related to this topic. 

Yes, agree.  In an editor using real tabs, size=4, the whitespace inconsistencies are irrelevant. 

>In future, I would do the white-space changing commits separate from the content changing commits.

Yes, agree, ... with myself.

> Obviously the whitespace changes need to be split out to a separate commit.

Yes, agree, and also think is not asking too much to have it separated.

----
> So i wrote a little utility, which converts leading spaces (for indentation) to tabs. What you see is the result.

OK, a tool like this likely the only way to enforce formatting consistently and should be proposed in a separate pull request or fb.net discussion.  It needs work though, for example, a few comments were removed, some leading spaces not converted to tabs, some lines with only white space, some lines with trailing whitespace.  And there's other whitespace inconsistencies in fbc besides only leading tabs.

General comments:
- The version V 0.5.0.1 is meaningless in the commit message; unless `ustring.bi` is going to be developed and released separately from `fbc`, then it makes sense; in which case version number should be part of the `ustring.bi` header.
- `contrib/swig/swig.exe` deleted? why? 
- commit of IDE specific `#compile` directives in `fbc.bas`?  These should not need to be committed.

PR #155, opened 2019-06-28
add ustring.bi

jklwn:

In Windows you should implement WinFBX as a recommended alternative, but "ustring.bi" must be included (after the "AFX..." includes !) nevertheless for proper working. "ustring.bi" contains code for
activating the USTRING feature in the compiler and macros for switching the "USTRING" keyword to José´s "AFX.CBstr".

The more there are new definitions, which might be useful when dealing with strings:

"DSTR" dynamic string type, = USTRING for UNICODE defined, STRING otherwise
"ZSTR" zero terminated string, = WSTRING for UNICODE defined, ZSTRING otherwise

This allows for writing code, which uses unicode string types for UNICODE defined and ascii string types otherwise the code will compile successfully in both cases.

JK


inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+
+  Public:
+    m_pBuffer AS UBYTE PTR                            'pointer to the buffer
+    m_BufferLen AS ulong                              'length in bytes of the current string in the buffer
+    u_size      as long = SizeOf(WSTRING)
+
+    DECLARE CONSTRUCTOR
+    DECLARE CONSTRUCTOR (BYVAL nChars AS ulong, BYVAL nCodePage AS ulong)
+    DECLARE CONSTRUCTOR (BYVAL pwszStr AS WSTRING PTR)
+    DECLARE CONSTRUCTOR (BYREF ansiStr AS STRING, BYVAL nCodePage AS ulong = 0)
```
jayrm: Probably should be `byref ansi as const string`.  Otherwise, it is not possible to initialize from a `const string`.  Applies to other declarations as well, as `const` not really used anywhere in this implementation.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+    DECLARE SUB Clear
+
+    DECLARE SUB Add (BYREF cws AS DWSTR)
+    DECLARE SUB Add (BYVAL pwszStr AS WSTRING PTR)
+    DECLARE SUB Add (BYREF ansiStr AS STRING, BYVAL nCodePage AS ulong = 0)
+
+    DECLARE OPERATOR [] (BYVAL nIndex AS ulong) byref AS USHORT
+
+    DECLARE OPERATOR CAST () BYREF AS WSTRING
+    DECLARE OPERATOR CAST () AS ANY PTR
```
jayrm: Doesn't this allow the `ustring` to be passed to any pointer type and effectively disables type checking?

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+    OPERATOR = zero                                   'return 0
+    exit operator
+  end if
+  
+  OPERATOR = *cast(USHORT ptr, m_pBuffer + (nIndex * u_size))
+END OPERATOR
+
+
+#else
+PRIVATE OPERATOR DWSTR.[] (BYVAL nIndex AS ulong) byref AS Ulong
```
jayrm: Feels like this conditional compilation could be combined with the one above, possibly using a `uchar` type.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+  this.Add(cws)
+END OPERATOR
+
+
+'***********************************************************************************************
+' Appends a string to the DWSTR
+'***********************************************************************************************
+
+
+PRIVATE OPERATOR DWSTR.+= (BYREF wszStr AS WSTRING)   'appends a wstring to the DWSTR
```
jayrm: This and many other parameters should be `const` otherwise won't take a `const` argument.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+' Properties
+'***********************************************************************************************
+
+
+PRIVATE PROPERTY DWSTR.Size() AS LONG
+  PROPERTY = u_size
+END PROPERTY
+
+PRIVATE PROPERTY DWSTR.Size (BYVAL nChars AS LONG)
+  IF nChars > -1 THEN u_size = nChars ELSE u_size = -1
```
jayrm: When would `u_size` ever need to change during runtime?  And if it did for some reason, zero (0), probably shouldn't be allowed.  Doesn't that break all the `\ u_size` math in length calculations?

tests/ustring/asc.bas
```diff
@@ -0,0 +1,144 @@
+#include "fbcunit.bi"
+#include "ustring.bi"
+
+
+#macro hCheckString( U, w )
+  r = 0                                               'reset flag
+
+  CU_ASSERT_EQUAL(u, w)
```
jayrm: This is duplicated in many (all?) of the tests.  Could be an include file instead.  Seems like the macro should generate the PASS/FAIL, rather than setting a flag.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+'***********************************************************************************************
+' DWSTR constructors
+'***********************************************************************************************
+
+
+PRIVATE CONSTRUCTOR DWSTR
+  this.ResizeBuffer(m_GrowSize)                       'Create the initial buffer
+END CONSTRUCTOR
+
+PRIVATE CONSTRUCTOR DWSTR (BYVAL nChars AS ulong, BYVAL nCodePage AS ulong)
```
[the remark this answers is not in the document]
JoseRoca: This class seems based in an old version of my CWSTR class. I changed the nCodePage parameter to a bClear parameter (meaning that the memory must be initialized or not) long time ago.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+
+PRIVATE CONSTRUCTOR DWSTR (BYVAL pwszStr AS WSTRING PTR)
+  IF pwszStr = 0 THEN
+     this.ResizeBuffer(m_GrowSize)                    'Create the initial buffer
+  ELSE
+     this.Add(pwszStr)                                'Add the passed WSTRING
+  END IF
+END CONSTRUCTOR
+
+PRIVATE CONSTRUCTOR DWSTR (BYREF ansiStr AS STRING, BYVAL nCodePage AS ulong = 0)
```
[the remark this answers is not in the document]
JoseRoca: It is used in my CWSTR class (Windows only) to allow to pass utf8 strings, but the calls to MultiByteToWideChar have been removed in ustring.bi taking away the purpose.

inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
...
+    DECLARE PROPERTY Size () AS LONG
+    DECLARE PROPERTY Size (BYVAL nValue AS LONG)
+    DECLARE PROPERTY GrowSize () AS LONG
+    DECLARE PROPERTY GrowSize (BYVAL nValue AS LONG)
+    DECLARE PROPERTY Capacity () AS ulong
+    DECLARE PROPERTY Capacity (BYVAL nValue AS ulong)
+    DECLARE PROPERTY SizeAlloc (BYVAL nChars AS ulong)
+    DECLARE PROPERTY SizeOf () AS ulong
+    DECLARE PROPERTY CodePage () AS ulong
+    DECLARE PROPERTY CodePage (BYVAL nCodePage AS ulong)
```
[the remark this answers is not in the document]
JoseRoca: I removed these properties long time ago.

jklwn:
General purpose:

There should be a default implementation for the new "EXTENDS WSTRING" feature. Ustring.bi could be a starting point for such an implementation. I would propose the key word "USTRING" for this dynamic wide string, because then it would be possible to use José´s type (depends on Windows) or the default type sharing the same name. Ustring.bi is an include file, so a user always has the choice to include it or to roll his own dynamic wide string type. By including José´s afx.cwstr this type becomes USTRING, if afx.cwstr isn´t included the type inside ustring.bi becomes USTRING (this is necessary for Linux e.g.).

jklwn:
Copyright and license:

Obviously my part of coding this type has been the smallest and José´s appears to be the biggest. I think we don´t have to discuss the disclaimer. All contributors declare this code to be freeware. So everbody is free to use it free of charge and for what ever he whishes. At the same time the copyright notice declares this code to be the intellectual property of the contributors. So nobody else can claim ownership, charge for it´s use or impose restrictions of use on it. I´m not a lawyer, but i think it should be clear now, that freebasic users can benefit from it free of charge and we (the contributors) want it to be and to stay this way. Jose, i know you read this, please clarify or contradict, if i´m wrong with this!

jklwn:
DSTR and ZSTR:

this allows for code, which can be compiled for unicode defined or not. 
>>>which currently has no meaning in fbc
I disagree, you already can write Windows code with unicode defined and the result will be different, if unicode isn´t defined. It´s a useful proposal, if a user doesn´t want it, he could (#)define DSTR/ZSTR before including ustring.bi. So a user always has a choice.

jklwn:
Regarding all other comments:

see it as a proposal, as a starting point. Feel free to improve it or fix it, where it doesn´t work. All i can say is, that i use this code for quite some time without experiencing problems. But this may be because i tested and used only a subset of what´s syntactically possible and allowed. I think the same applies to José´s implementation, which is in use for some years now. But we cannot take his type as is, because it depends on Windows and we need a Linux counterpart too. Therefore (apart from some minor corrections and additions, which fixed or improved José´s version) i reduced his version to a working minimum. There is room for improvement, sure - make the best of it. My goal is not to see my name or my initials in some copyright notes, my goal is improvment! 

Let´s do it just like with what ended up as "EXTENDS WSTRING": i do (or in this case basically present) some raw and basic work and you do the fine tuning and take it to a level, which can be added to master.

JoseRoca:
> Copyright and license:
> 
> Obviously my part of coding this type has been the smallest and José´s appears to be the biggest. I think we don´t have to discuss the disclaimer. All contributors declare this code to be freeware. So everbody is free to use it free of charge and for what ever he whishes. At the same time the copyright notice declares this code to be the intellectual property of the contributors. So nobody else can claim ownership, charge for it´s use or impose restrictions of use on it. I´m not a lawyer, but i think it should be clear now, that freebasic users can benefit from it free of charge and we (the contributors) want it to be and to stay this way. Jose, i know you read this, please clarify or contradict, if i´m wrong with this!

For the purpose of writing a dynamic wstring class to be distributed with FreeBasic, the license used with the other FreeBasic .bi files if fine with me, but then I don't understand all this mess with jk.DWSTR, afx.CWSTR, USTRING, DSTR, ZSTR and \_\_FB_DOS_\_\ (who is going to use this class with DOS?).

jklwn:
José,

your afx.CWSTR is for Windows only. I know you are not interested in Linux, but FreeBASIC is multi-platform. So we need something for Linux (and DOS). As far as i know, DOS doesn´t have wide strings, so i redefine USTRING to STRING. You cannot work with wide strings then, but the code will compile and run with ANSI strings. For Linux users USTRING will be a wide string type they can use, if they want. Linux can deal with UTF-8 encoded wide strings, so most Linux user go that way. But again existing code is compilable and runs in Linux too (as long as you don´t use the Windows API of course). For Linux the reduced type in ustring.bi (jk.DWSTR, FB.DWSTR, whatever) becomes USTRING. In Windows this type is the default too, except your type is included before in cluding ustring.bi. Then afx.CWSTR becomes USTRING. That is, we can have the key word "USTRING" everywhere, even if the underlying, platform specific implementation differs.  

DSTR/ZSTR is for "dynamic string" and "zero terminated string". Using these two defines makes the repeated use of #if ... #else ... #endif blocks for string/ustring and zstring/wstring obsolete. It´s an offer not a must, everone can still do it as he wishes.

JoseRoca:
I can understand making changes to make it Linux compatible, but not the rest. It makes no sense to me to make the class to work with ansi when FB already has ansi data types. If it does not compile when it is being misused, I think that this is a good warning that they are making an idiocy.

> That is, we can have the key word "USTRING" everywhere, even if the underlying, platform specific implementation differs.

For total compability, it is absolutely necessary that the interfaces are the same. Any difference in methods/properties names, parameters, behaviour, etc., will break it.

jklwn:
The class actually doesn´t work with ANSI, i just define USTRING as STRING in this case. Total compatibility is not possible with different classes, otherwise the classes weren´t different. What i want to acceive, is a platform independent compatibility with FreeBASIC statements. Using class specific methods or properties will break it, that is true. But when using USTRING just like any other string type (not calling methods or properties directly) it should be compatible with all platforms. 

If i´m not mistaken WSTRING is redefined for DOS, because DOS cannot process wide strings. Why not have it for USTRING in a similar way?

jayrm:
@jklwn, fbc in DOS defines WSTRING having a size of 1 byte.  Compare to 2 bytes on Windows and 4 bytes on Linux.   For consistent behaviour across all platforms, there's no reason to define ustring as something completely different on DOS.  "extends" syntax is still valid on DOS.  How useful is it?  Leave that up to the user.

As mentioned by @JoseRoca, just leave  jk.DWSTR, afx.CWSTR, USTRING, DSTR, ZSTR and all that mess up to the user.  There's no need to enforce the usage as it is all just mapping of a name to a type.

My preference would be to limit the number of global symbols to the minimum, something like:
```
namespace fb
	type VarLenWstring extends Wstring
		'' ...
	end type
end namespace
```

And that's it.  Leave it to the user to `#define` or `type` it as something else.  Basically, go for the minimal solution.

jayrm:
For the licensing, it is a technicality, but an important one: the copyright holder establishes the license.  And as the copyright holder you can change the license anytime you want to suit your own purposes.  As a developer (and a user), what I am looking for is some guarantee that I can continue to use and modify the code even if you still own the copyright.

For this I would use the following (as I have done in fbcunit code):
```
''  fbcunit - FreeBASIC Compiler Unit Testing Component
''	Copyright (C) 2017-2019 Jeffery R. Marshall (coder[at]execulink[dot]com)
''
''  License: GNU Lesser General Public License 
''           version 2.1 (or any later version) plus
''           linking exception, see license.txt
```

There's all kinds of opensource licenses out there, but this is same one that is used with FreeBASIC rtlib.  Most of the headers just use the same license as the original source, since it has been translated.  Then there is a few headers that may have this notice:
```
'' NOTICE: This file is part of the FreeBASIC Compiler package and can't
''         be included in other distributions without authorization.
```
Though it might be implied that the FreeBASIC Development Team has the copyright on those files since it is not explicitly stated in a copyright notice.

Bottom line ... it's just easiest if the copyright holder publishes the source with a compatible license.

rversteegen:
I agree that simply declaring the code is "Freeware" isn't actually commonly understood to give all the rights needed. But then jklwn wrote:

>  All contributors declare this code to be freeware. So everbody is free to use it free of charge and for what ever he whishes. At the same time the copyright notice declares this code to be the intellectual property of the contributors. So nobody else can claim ownership, charge for it´s use or impose restrictions of use on it. I´m not a lawyer, but i think it should be clear now

I think what you mean is that you reserve moral rights (e.g. that other people can't claim ownership of your code or claim rights over it) and dismiss copyright rights over it.
On the other hand if I take what you said literally, that you want to restrict anyone from selling it, that would make these headers NOT open source/free software (as defined by OSI, FSF or Debian)! That makes it incompatible with GPL and LGPL, for example. I don't think that's what you meant.

So please lets have a clear statement.

I realise this is a bit off-topic, but...
>'' NOTICE: This file is part of the FreeBASIC Compiler package and can't
>''         be included in other distributions without authorization.
This is not a open-source nor free source license, because it doesn't fully permit redistribution. I think these notices are terrible and really ought to be replaced.

In fact where this notice has been slapped on headers translated from C it might be in violation of the original license terms. Luckily, in those case your restriction against inclusion in other distributions is likely unenforceable because my understanding (IANAL) is that header files which have simply been run through a translation tool like fbfrog do not contain a contribution significant enough to be copyrightable.

rversteegen:
To add to that last point, whether header files containing only declarations (eg SWIG translated headers) are even copyrightable in the first place is another question. Hopefully they aren't (most people would say they aren't), but see Oracle vs Google.

jklwn3:
IANAL too, so if the the license you propse serves the purpose, let´s take it. José agrees too, as i see it
<<< ... the license used with the other FreeBasic .bi files if fine with me ...

Ustring: 
The idea behind redefining it as USTRING, is that this makes code more re-usable. If someone wants to implement his own "extends wstring" type, the only thing, which must be changed in existing code is the USTRING definition, no need to change names in maybe hundreds of places. Obviously someone including ustring.bi wants to use this type, so defining it as USTRING makes sense to me.  I could drop "ZSTR" and "DSTR" as well as making José´s CWstr the default USTRING if present. José could also add an USTRING definiton to his file, if he wants. 

The type in ustring.bi won´t work for sizoef(wstring) = 1 (DOS), it´s useless in DOS anyway. Therefore i define USTRING as STRING in this case, which makes existing code (implementing USTRING) at least partially (no real unicode strings) usable.

The more, you have seen my ideas for extending array and string features. I would like to make it work for USTRINGs too. These extensions won´t work for different "extends wstring" type names, i need a common name. If a different (extends wstring) implementation complies to the defaults set by ustring.bi, it can work with these extensions as well. You just have to define this type as USTRING.

Defining USTRING doesn´t do any harm, everyone can still do as he pleases, he can use it or not. But including ustring.bi should include defining USTRING. You can always (#)undef it, if you don´t like it. 

I think, if  the average user will ask himself: "why didn´t they include it, why do i have to do it myself (#define USTRING ...), if it´s the default (or recommended) way anyway".

Please think about it!

JK

jklwn:
Will make a new one ...

jayrm:
OK, so this PR is abandoned and recreated in #161

PR #156, opened 2019-06-28
White space

jklwn:
This tries to make indentation consistent by converting leading spaces to tabs in the compiler and RTL sources. Comments not beginning with two single quotes ('') in .bas and .bi files are removed just as comments in .c and .h files not enclosed by "/* .... */".

JK


inc/ustring.bi
```diff
@@ -0,0 +1,681 @@
+
```
jayrm: If this PR is for whitespace, then I do not expect to see a new file inc/ustring.bi added.  ustring.bi is also added in PR #155 and the 2 versions are different.  Maybe this was added by mistake?

src/compiler/ast-node-bop.bas
```diff
@@ -1539,7 +1539,7 @@ function astNewBOP _
 			return rtlMathLongintMOD( dtype, l, ldtype, r, rdtype )
 		end select
 
-	' Trap ANDALSO, ORELSE, handle floats, and convert to IIF
```
jayrm: Here, and in many other files, comments with single quote are stripped.  My preference is 2 single quotes to mark a comment, but that does not mean single quoted comments should be removed.

src/compiler/stabs.bi
```diff
@@ -1,40 +1,40 @@
-const STAB_TYPE_GSYM 	= 32 'Global symbol
```
jayrm: Comments are removed here and should not be.  Should just leave it as is.

tests/ustring/asc.bas
```diff
@@ -0,0 +1,144 @@
+#include "fbcunit.bi"
```
jayrm: new files in tests/ustring/* should not be part of a whitespace only commit

src/rtlib/unix/sys_dylib.c
```diff
@@ -17,11 +17,11 @@ FBCALL void *fb_DylibLoad( FBSTRING *library )
 							  "./lib%s.so",
 							  NULL };
 
-	// Just in case the shared lib is an FB one, temporarily reset the
```
jayrm: Newer C compilers should accept `// comments' this.  For portability, all C compilers will accept `/* comments */'.  Regardless, the comments should not be removed.

jayrm:
As mentioned before, in an editor that supports real tabs, spacing=4, the whitespace issue is quite irrelevant.  But if you really feel that the whitespace and formatting of fbc source must be addressed, then consider publishing the filter/sanitizer.  There's really no point in addressing whitespace within the code base without a reliable automatic method (program) to apply the rules.

jklwn:
<<<But if you really feel that the whitespace and formatting of fbc source must be addressed, then consider publishing the filter/sanitizer. 

Maybe i will do that.

PR #158, opened 2019-07-03
fix passed z/wstring array bug

jklwn3:
This fixes a bug mentioned in https://sourceforge.net/p/fbc/bugs/650/. Passed z/wstring arrays didn´t work properly. The more such arrays are now accepted without "*..." term in procedure definitions (sub foo (array() as zstring)), which is meaningless anyway. Sizeof still doesn´t work as expected

sub foo( array() as zstring )
    print array(0), sizeof( array(0) )
end sub

dim array(0 to 0) as zstring * 32
array(0) = "hello"

print array(0), sizeof( array(0) )
foo( array() )

- added: SIZEOF now works as well, assignment will follow
- added: assignment works, maybe must modify file put/get functions too
- added: data, file functions + mem swap

This pull request is ready now.

### DIFF
--- a/src/compiler/parser-decl-symbtype.bas
+++ b/src/compiler/parser-decl-symbtype.bas
@@ -141,8 +141,6 @@ private function cSymbolTypeFuncPtr( byval is_func as integer ) as FBSYMBOL ptr
 	attrib = 0
 	subtype = NULL
 
-	' no need for naked check here, naked only effects the way a function
-	' is emitted, not the way it's called
 
 	'' mode
 	mode = cProcCallingConv( )
@@ -231,8 +229,8 @@ sub AmbigiousSizeofInfo.maybeWarn( byval tk as integer, byval refers_to_type as 
 	'' Don't warn if it's a type and a var of that type, because then the
 	'' len() or sizeof() would return the same value in either case.
 	if( symbIsVar( nontype ) and _
-	    (symbGetType( nontype ) = FB_DATATYPE_STRUCT) and _
-	    (nontype->subtype = typ) ) then
+		(symbGetType( nontype ) = FB_DATATYPE_STRUCT) and _
+		(nontype->subtype = typ) ) then
 		exit sub
 	end if
 
@@ -290,8 +288,8 @@ function cTypeOrExpression _
 
 	'' Token followed by an operator except '*' / '<'?
 	if( (lexGetLookAheadClass( 1 ) = FB_TKCLASS_OPERATOR) andalso _
-	    (lexGetLookAhead( 1 ) <> CHAR_TIMES) andalso _
-	    (lexGetLookAhead( 1 ) <> FB_TK_LT) ) then
+		(lexGetLookAhead( 1 ) <> CHAR_TIMES) andalso _
+		(lexGetLookAhead( 1 ) <> FB_TK_LT) ) then
 		maybe_type = FALSE
 	else
 		'' Check for some non-operator tokens
@@ -439,14 +437,14 @@ private function cMangleModifier _
 						dtype = typeSetMangleDt( dtype, FB_DATATYPE_UINT )
 						function = TRUE
 					case else
-						errReport( FB_ERRMSG_SYNTAXERROR )	
+						errReport( FB_ERRMSG_SYNTAXERROR )    
 					end select
 				else
 					select case dtype
 					case FB_DATATYPE_LONG
 					case FB_DATATYPE_ULONG
 					case else
-						errReport( FB_ERRMSG_SYNTAXERROR )	
+						errReport( FB_ERRMSG_SYNTAXERROR )    
 					end select
 				end if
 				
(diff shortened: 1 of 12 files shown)


src/compiler/parser.bi
```diff
@@ -315,7 +315,8 @@ declare function cSymbolType _
 		byref subtype as FBSYMBOL ptr, _
 		byref lgt as longint = 0, _
 		byref is_fixlenstr as integer = FALSE, _
-		byval options as FB_SYMBTYPEOPT = FB_SYMBTYPEOPT_DEFAULT _
+		byval options as FB_SYMBTYPEOPT = FB_SYMBTYPEOPT_DEFAULT, _
+		byval mode as INTEGER = 0 _
```
jayrm: I've noticed this in several files from your repos, that there is trailing space added at the end of the line.  Why?

src/compiler/parser-decl-symbtype.bas
```diff
@@ -141,8 +141,6 @@ private function cSymbolTypeFuncPtr( byval is_func as integer ) as FBSYMBOL ptr
 	attrib = 0
 	subtype = NULL
 
-	' no need for naked check here, naked only effects the way a function
```
jayrm: Comments removed.  Why?

src/compiler/parser-decl-symbtype.bas
```diff
@@ -505,25 +503,26 @@ end function
 
 '':::::
 ''SymbolType      =   CONST? UNSIGNED? (
-''				      ANY
```
jayrm: OK, so tabs have been replaced with spaces for alignment.   I understand.  Except, the comment is not aligned.  Even though these are whitespace changes, still need to review and check that they are correct.  What's the point of the change if needs to changed again?

jayrm:
I don't want to focus on the whitespace, but it is seriously a major distraction from the changes you want to contribute.

jklwn3:
I know those whitespace changes are a nuisance. We should discuss this topic in private, i will drop you a mail soon (work is killing me at the moment, so maybe at the weekend). 

Personally i prefer WinMerge over TortoiseGitMerge, because i can merge folders file by file and i can ignore whitespace changes when diffing and merging.  

I think you review each changed line anyway, so hand-merging vs automatic merging wouldn´t be so much more work. The advantage is that you don´t have to accept the commit history as is, by hand-merging you can skip commits of a pull request and add it as one or more commits of your choice to master.

JK

jayrm:
So, the main issue with the whitespace, is that even though options exists for github & git to ignore whitespace changes, if we ignore whitespace, that works and we see only the code changes, however, if the whitespace changes are actually wrong, we miss it.  Entire files could be indented incorrectly and we wouldn't notice if ignoring the whitespace changes.  Basically, an editor or IDE should not be forcing one style of indentation or the other.  Some files (like makefiles) actually use both types of indentation spaces and tabs have different meaning.  Editor and IDE should only rewrite lines actually edited.

jayrm:
The main issue I have with this "fix" is that it solves a compile-time problem with run-time expressions.  This means that values that should be compile-time constant that are currently computed are incorrect (that's the real bug), and are instead solved with code evaluated at run-time.  fbc is pretty good at optimizing constant expressions.  This fix takes that possibility away, so this is wrong solution.

With this "fix" we still have compiler incorrectly identifying array types, which I think is the root of the problem.  As should be expected, we need to go deeper to find the real cause.

For example,
```
sub foo( array() as zstring * 32 )
	#print typeof( array(0) )
end sub
'' Output:
'' ZSTRING * 1
```

jklwn:
The root of the problem is, that currently there is no way of passing the size of z/wstrings. Only the data type is passed, which is a problem for z/wstrings. I circumvent this for arrays by taking the element_len member of the array descriptor (which holds the size). Fixing this for good, would require to change how you pass z/wstrings in general, i.e. adding a method of passing the size too.


JK

jklwn:
Maybe i should explain a bit more, why i did it this way and not the way you propose:

Consider the following code:

```
dim a(1) as zstring * 16 
dim b(1) as zstring * 32 
a(1) = "Hello from a"
b(1) = "Hello from b"


sub show(c() as zstring)

  print c(1)

end sub

show(a())
show(b())


sleep
end
```

This raises an error ("Expected pointer, found ')'"). Adding e.g.  "* 16 to the sub´s header makes it run, but there is no visible output. This is the current situation.

With my fix applied the output is:

> Hello from a
> Hello from b
 
just as expected.

If you apply a compile time fix the resulting procedure will accept only arrays of zstrings of a certain (predefined, e.g. zstring * 16) size. Arrays of zstrings of a different size (e.g. zstring * 32) will raise an error. This is not what you want!

Take a look at the following code:

```
dim a as zstring * 16 = "Hello from a"
dim b as zstring * 32 = "Hello from b"


sub show(c as zstring)

  print c

end sub


show(a)
show(b)


sleep
end
```
You can pass a zstrings of different size to a procedure. Why shouldn´t this be allowed for an ARRAY of zstrings? In case of a single zstring you can get away without knowing the size (as long as you don´t exceed the bounds of the allocated buffer). In case of an array you don´t get away without the size. In order for a procedure to be able to receive arrays of zstrings of different sizes (which should be possible for consistency) you must handle this at run time. 

My approach solves the problem for arrays, it doesn´t solve it for single zstrings. And i don´t know of good (compiler based) solution for passed (byref) zstrings. The risk of overwriting and thus corrupting memory by exceeding the bounds of the memory allocated for a zstring exists in exactly the same way in PowerBASIC and in C too. As a user you just must take care your buffer is big enough (have another parameter for the buffers size, whatever, ...).

PR #161, opened 2019-07-19
added ustring.bi + tests

jklwn:

JK


jklwn:
IANAL too, so if the the license you propose serves the purpose, let´s take it. José agrees too, as i see it
<<< ... the license used with the other FreeBasic .bi files if fine with me ...

Ustring:
The idea behind redefining it as USTRING, is that this makes code more re-usable. If someone wants to implement his own "extends wstring" type, the only thing, which must be changed in existing code is the USTRING definition, no need to change names in maybe hundreds of places. Obviously someone including ustring.bi wants to use this type, so defining it as USTRING makes sense to me. I could drop "ZSTR" and "DSTR" as well as making José´s CWstr the default USTRING if present. José could also add an USTRING definition to his file, if he wants.

The type in ustring.bi won´t work for sizeof(wstring) = 1 (DOS), it´s useless in DOS anyway. Therefore i define USTRING as STRING in this case, which makes existing code (implementing USTRING) at least partially (no real unicode strings) usable.

The more, you have seen my ideas for extending array and string features. I would like to make it work for USTRINGs too. These extensions won´t work for different "extends wstring" type names, i need a common name. If a different (extends wstring) implementation complies to the defaults set by ustring.bi, it can work with these extensions as well. You just have to define this type as USTRING.

Defining USTRING doesn´t do any harm, everyone can still do as he pleases, he can use it or not. But including ustring.bi should include defining USTRING. You can always (#)undef it, if you don´t like it.

I think, if the average user will ask himself: "why didn´t they include it, why do i have to do it myself (#define USTRING ...), if it´s the default (or recommended) way anyway".

Please think about it!

JK

jklwn:
I took out "codepage", because it is not necessary in this implementation. 

Regarding "declare operator cast() byref as wstring" (without "CONST") bypassing the type´s internal logic: there should be no problems with "L/RSET", "MID" (statement), because the length doesn´t change with these statments. "SWAP" as you coded it is safe too. So i opted for the version without the CONST qualifier as a default.

JK

jayrm:
Many of my review comments of #155 still apply and have not been addressed.  Why must I review the same code a second time?

I think it best practice to deliver the absolute minimal code that satisfies the requirement.  It's much easier (and user friendly) to add later if it's really needed than to change or remove code.

jklwn:
No, it´s not the same code! I addressed your major points of critics and changed them as you proposed (License, codepage, removed crt/mem.bi dependecy, making José´s type the default, if present, i tested the "CONST" problem - it works and i posted a corresponding reply).

As i see it, the rest of it is more or less a matter of taste or style. This code has been developed by several capable programmers (i´m not speaking of myself!), it has been in use for several years. So it can be assumed that it most probably runs bug free now. My part has been by far the smallest, but i´m the one trying to make it an official part of the distribution, so that the community can benefit from this fine work.

It serves the purpose, and sometimes good enough is good enough. Every piece of code can be written in many different ways. And sometimes making working code even smarter makes it even worse. Accepting it as it is, doesn´t mean you cannot add, change, improve whatever you want later on.

What more do you want? I present working and tested code, i wrote and ran and added tests for it. I said more than once: feel free to change whatever you think still must be changed to meet your needs.

JK

jayrm:
OK, got it. Thanks.  @jklwn, I know you have put a great amount of effort in to this feature, and you have addressed many of the comments I made in #155.  So, I agree to take it the rest of the way and soon we will get this merged in.  I will force push changes to this PR, but it will be your original contributions plus my modifications.  My goal is to have most concise code and fewest new symbol names. True, I may be biased on personal taste and style for some points, but on other points, like removing duplicated code, I am confident other developers would agree that removing the duplicated code is best goal.  Thanks again.

jklwn:
Thanks Jeff, it passes all my Windows tests (not contained in the test suite) i wrote while working on it!

JK

jayrm:
Yes, I'd like to finish this PR off; it's fairly close to the top of my todo list to get this finished and merged in.

Originally, the 'ustring.bi' was pretty much copy+paste from Jose's WinFBX followed by some hack+slash.  It might be in use for quite a while so I think it is worth taking it's development as far as is reasonably possible before formally adopting it's use, rather than just throwing it out there because it "works".

I think need to explore what this thing is and why we are adding it:
- probably a long time before var-len wstring is natively supported by fbc
- would be helpful to users to have an equivalent even if implemented as an #include
- would help development in other areas of var-len wstring support
- we are not building a framework or data type system

What we want here is a place holder that will work a lot like a native var-len wstring would work.  Which for symmetry, should likely have many similar characteristics as already existing STRING:
- minimal field: data, length, size
- null descriptor if no string is allocated
- fixed growth rate (STRING uses 12.5%)
- similar operations allowed or restricted as STRING
- declaration separate from implementation (especially if supporting code is large)

As is 'ustring.bi' has these extra characteristics:
- pre-allocates 260 bytes, even for empty strings
- growth rate field is stored with every string
- implicit conversion construction like 'dim x as ustring = 1'
- executable code in the include (duplicates code in multi-module compilations)
- the actual type is FB_USTRING.DWSTR, but has `#define ustring FB_USTRING.DWSTR` indirection
- overloads CAST() AS ANY PTR allowing ustring to be passed to any pointer type without warning/error.
- no awareness of CONST qualifiers

I'd kind of like to get 'ustring.bi' down to just the minimum needed.  Simple use, easy to maintain & document, etc.

jklwn:
Jeff, i removed some unnecessary things. There is a new constructor, i plan to use for two things:
- don´t initialize anything, this skips default constructing when used in the RTL 
- set the initial size of u_size (former m_growsize) other than the default for special situations (e.g make repeated concatenations faster)

There is a problem i don´t understand in line 318. Increasing u_size by 12.5% (u_size * 0.125) is not enough for the SWAP statement in my test code. It needs to be at least 43 * UCHAR_SIZE - strange.

There will be another commit for the overloaded LEFT/RIGHT statements. Currently embedded nulls are unsupported (LEFT stops at the first embedded null, even if the actual string is longer). I want to change that, because the new string processing functions will support embedded nulls for ustrings.

jklwn:
Trying to get rid of the constructor for longint and double fails, because i cannot overload the "&" several times. That is, i can, but only the most recent operator (the last in code) takes effect (see line # 382 and following). 

I still don´t understand the need for at least 43 added UCHARS in line 301 for SWAP to work.

It would be nicer, if we could get rid of the constructor for longint and double, because then my special constructor could be a long (which looks more straightforeward) instead of an additional type (which btw. works, but is kind of ugly - on the other hand not so easy to make it by mistake though)

jklwn:
Changing the conversion method in PRIVATE SUB DWSTR.Add (BYREF ansiStr AS STRING) allows for embedded nulls in incoming string. Previous method stopped at the first null.

jklwn:
Added operator cast as STRING, convert from USTRING to STRING with embedded nulls, intrinsic conversion stops at first nulls. While this part works, there are still problems with parameter conversion and return type conversion. Therefore the added code is currently outcommented.

jklwn:
I think, now it acts like STRING except for embedded nulls. Constructing it with an ULONG sets the size of the initial buffer, = 0 -> no allocation takes place (this skips default allocation for special cases), > 0 allocate this number of characters (if you know, that you need a larger buffer for concatenations - speeds this up).

jklwn:
Added a new statement "SHIFT", which is a wrapper for C memmove. Might be useful in FB too. I need to have a function moving memory with out adding a terminating null, the previously implemented  fb_hStrCopy did this.

PR #169, opened 2019-08-26
fbc: add array descriptor 'flags' field

jayrm:
- update ./fbc-int/array.bi with new flags field
- flags field records if array is fixed dimension, fixed length, and max number of dimensions available in dimTb()
- const FBARRAY_FLAGS_DIMENSIONS = &h0000000f    '' number of entries allocated in dimTb()
- const FBARRAY_FLAGS_FIXED_DIM  = &h00000010    '' array has fixed number of dimensions
- const FBARRAY_FLAGS_FIXED_LEN  = &h00000020    '' array points to fixed-length memory
- const FBARRAY_FLAGS_RESERVED   = &hffffffc0    '' reserved, do not use

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,9 +1,11 @@
 Version 1.08.0
 
 [changed]
+- array descriptor contains new 'flags' field.  Careful, breaks binary compatibility plus chicken-egg problem to build fbc.
 
 [added]
 - extern "rtlib": respects the parent namespace, uses default fb calling convention and C style name mangling
+- array descriptor 'flags' field to track fixed length, fixed dimension, dimTb() size.
 
 [fixed]
 
--- a/inc/fbc-int/array.bi
+++ b/inc/fbc-int/array.bi
@@ -15,17 +15,23 @@ namespace FBC
 		dim as integer ubound        '' dimension upper bound
 	end type
 
+	const FBARRAY_FLAGS_DIMENSIONS = &h0000000f    '' number of entries allocated in dimTb()
+	const FBARRAY_FLAGS_FIXED_DIM  = &h00000010    '' array has fixed number of dimensions
+	const FBARRAY_FLAGS_FIXED_LEN  = &h00000020    '' array points to fixed-length memory
+	const FBARRAY_FLAGS_RESERVED   = &hffffffc0    '' reserved, do not use
+
 	type FBARRAY
 		dim as any ptr index_ptr     '' @array(0, 0, 0, ... )
 		dim as any ptr base_ptr      '' start of memory at array lowest bounds
 		dim as uinteger size         '' byte size of allocated contents
 		dim as uinteger element_len  '' byte size of single element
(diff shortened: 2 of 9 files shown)


rversteegen:
What is the "chicken-egg problem to build fbc"? What will be the minimum required version of fbc to build fbc after this?

Indicating whether an array is resizable or not will be very nice if this means that you are also going to add checks to the rtlib to throw an error if you attempt to redim an array of a fixed size or limited number of dims, instead of crashing!

jklwn:
As i see it, you must first compile the RTL and exchange libfb(mt).a, etc., in your build environment Then you can compile the new compiler sources linked to the new RTL. This result´s in fbc 1.08, which should be able to recompile itself. Jeff, is this correct?

May i use
const FBARRAY_FLAGS_ATTACHED = &h00000040 '' array overlays existing memory
then?

JK

jklwn:
We cross posted, what a coincidence!

Shouldn´t there be problems after your step 1, because the offset to the dimTB is different (one uinteger more) in the compiler´s code from the offset in the older RTL? Functions like L/UBOUND() cannot work properly with a wrong structure offset- just wondering

JK

rversteegen:
> These are normal steps for building

OK, good, so steps 2 and 3 which you gave are optional (though obviously ensuring fbc 1.08 can build itself must be done). I never bother to compile `fbc` twice, I still compile FB from git with FB 1.05 - is that bad? My impression was that you have a policy of ensuring that `fbc` can be compiled/linked by quite old versions of FB. Contrary to what JK wrote, I can't see how internal implementation details of the runtime library that `fbc` is linked to affects the programs it produces.

rversteegen:
I guess I misread @jklwn: to answer your question, in each of those 3 steps, `fbc` is always compiled by a version of `fbc` matching the runtime version it was linked to.

jklwn:
Seems it works both ways (Jeff´s method and mine). I can build V 1.08 in 32 and 64 bit Windows. Both versions can recompile themselves, i tried 5 times in a row. 

The only problem i have is the makefile. I had to use an older version (1.05) to make it compile the RTL. I don´t use the makefile for the compiler.

JK

rversteegen:
Thanks for the explanation!

jayrm:
The force push was required to rebase after #172 which allows the use of `extern "rtlib"` for declarations.

The declaration for `fb_ArrayGetDesc()` is now moved inside the FBC namespace.  And now has two ways to call the procedure:
- `FBC.ArrayDescriptorPtr( array() as any ) as FBC.FBARRAY ptr`
- `FBC.ArrayConstDescriptorPtr( array() as const any ) as FBC.FBARRAY const ptr`
- both entry points are supported by same run time `fb_ArrayGetDesc()` function,  Only difference is the data type returned to the compiler for compile time data type checks (i.e. with const).

Normally, I would spend a little more time writing tests, however, I want to allow development to keep on going, and I think the tests that are written prove the immediate use, so should be acceptable to merge.

PR #171, opened 2019-08-29
Add FreeBSD x86_64 'make bootstrap-dist' target

vilhelmgray:


dkl:
Looks good. So, did you by any chance find out whether the "make bootstrap" produces a working fbc with this?

rversteegen:
What about freebsd-x86, does anyone use it?

vilhelmgray:
> Looks good. So, did you by any chance find out whether the "make bootstrap" produces a working fbc with this?

Yes, I was able to compile a working fbc on FreeBSD after running `make bootstrap-dist` on Linux to produce the bootstrap code.

vilhelmgray:
> What about freebsd-x86, does anyone use it?

To add this would we simply need to use `-target freebsd-x86` instead of `-target freebsd-86_64`. I can update this pull request to add it if you believe it's prudent -- [FreeBSD i386 has a support tier of Tier 1](https://www.freebsd.org/platforms/i386.html).

jayrm:
I was able to bootstrap fbc on FreeBSD-12.0-i386 using Ubuntu 14.04-i386, so I have added `-target freebsd-x86` as well to this PR.

I only tried a couple of simple programs with the newly built compiler, since running the test suite requires some additional work in `./inc/crt/` and maybe other places, which is more than I want to get in to now.  Regardless, both FreeBSD targets seem to valid for boot strapping and can be added to the future boot strap release packages.

PR #180, opened 2019-09-20
make file/path related functions work for wide strings in Windows

jklwn4:
This PR tries to make file related functions in Windows work with wide strings. There will be one commit for each statement.

JK


jklwn:
_wfopen doesn´t exist in Linux, so i must restrict it´s use to Windows. Maybe tomorrow ...

rversteegen:
The use of `#ifdef __FB_WIN32__` in rtl-file.bas is wrong, because it breaks cross-compiling. Instead you need to do a runtime check of `env.clopt.target`. Search for `FB_COMPTARGET_WIN32` for examples.

What does the comment
`	dim shared as FB_RTL_PROCDEF funcdata( 0 to 74 ) = _                    'jk-file  71 + 3 -> (71 + 6)`
mean?

There are a couple places you wrote `fname_len*2` or `fname_len * 2 + 2` which I think ought to be `fname_len * sizeof(FB_WCHAR)` etc, since those functions aren't in win32/*.c files, so they might theoretically be compiled on non-Windows OSes.

To reduce code duplication I think you shouldn't duplicate `hCheckBOM_W` and `hWriteBOM_W`, but instead make the existing functions in src/rtlib/dev_file_encod_open.c non-static and call them from dev_file_encod_open_w.c.
Duplication is unfortunately hard to avoid -- but the code can be deduplicated as cleanup at a later time, rather than demanding it's perfect now.

`#if defined( HOST_DOS ) || defined( HOST_MINGW ) || defined( HOST_XBOX )`
(in src/rtlib/file_hconvpath_w.c) isn't right, for example it ignores cygwin. I see you copied rc/rtlib/file_hconvpath.c, which is where the error came from. I think it should be
`#if defined( HOST_DOS ) || defined( HOST_WIN32 ) || defined( HOST_XBOX )`

jklwn4:
Thanks for commenting!

Why is #ifdef __FB_WIN32__ wrong, i can see it in other places too? I wanted to make it run in Windows in first place and later optimize it, where possible and reasonable. 

The comment is a temporary reminder for me, not to get out of sync with functions, there will be more functions to add. It will be removed later.

Yes, i should make it sizeof(FB_WCHAR).

I´m a Windows guy and my knowledge of cywin, XBOX and the like is very, very limited if not existing at all. So i try to make it work for Windows specifically, compatibility for other targets must be added by someone else.

jklwn4:
Ok, a first step is made. It works in Windows and it compiles and passes tests here.

Time to make a decision: make it work both ways ANSI and Unicode (this ensures Windows 95 compatibility) or handle Windows as wide string only (minimum requirement Windows XP) ?

Jeff, what do you think?

JK

rversteegen:
As I said, it breaks cross-compiling. Maybe you aren't aware that FB can cross-compile, using the -target commandline argument. If you write
```
#ifdef __FB_WIN32__ 
           f = PROCLOOKUP( FILEOPEN_W )
```
then which rtl proc is used depends on the platform fbc was compiled for rather than the platform it's compiling for.

What did you change when you force-pushed? I don't see any differences.

BTW Win XP isn't the minimum requirement if using unicode - that would be Win NT 3.1 (from 1993!), or Win 2000 if considering just consumer Windows versions.

rversteegen:
Oh, and to answer your question: if you look at the existing places __FB_WIN32__ is used, they are all things that depend on the host rather than the target, for example processing/building paths to files on the host.

jklwn4:
> Maybe you aren't aware that FB can cross-compile, using the -target commandline argument

Indeed, this was a misconception. Thanks for telling me! I had to amend the first commit several times until Travis was satisfied.

rversteegen:
Check `FB_COMPTARGET_CYGWIN` too, not just `env.clopt.target = FB_COMPTARGET_WIN32`.

And maybe xbox too? Eh, I think no one cares about xbox anymore, it's a long dead FB target, right?

I've never tried FB on cygwin, I wonder whether it still works. I suspect it doesn't, since it looks like checks for `FB_COMPTARGET_CYGWIN` might be missing in a few places in the compiler.

jayrm:
Seems like this should follow the pattern of the quirk string functions.  We just did this kind of thing in `extends w|zstring` features in PR #150.
```
astTryOvlStringCONV( filename )

if( fencoding = NULL ) then
	if( astGetDataType( filename ) <> FB_DATATYPE_WCHAR ) then
		proc = astNewCALL( PROCLOOKUP( FILEOPEN ) )
	else
		proc = astNewCALL( PROCLOOKUP( FILEOPEN_W ) )
	end if
else
	if( astGetDataType( filename ) <> FB_DATATYPE_WCHAR ) then
		proc = astNewCALL( PROCLOOKUP( FILEOPEN_ENCOD ) )
	else
		proc = astNewCALL( PROCLOOKUP( FILEOPEN_ENCOD_W ) )
	end if
end if
```

jklwn4:
Heureka! This time it was really hard to satisfy Travis...

jklwn4:
A first step is made, everything seems to work fine so far. I made the Windows version internally work with wide strings only. That is, you now can pass or return ANSI and wide strings to or from all file/path related functions. 

All new RTL files have been named by appending "_w" to already existing files. These files contain cloned wide string versions of the existing function chains. It´s "_w" instead of "_wstr" in order to easily keep apart, what was already there ("_wstr") and what has been added ("_w"). This should be changed later. 

There are still inconsistencies and probably unnecessary code sections, which must be removed, but basically it works.

JK

jklwn4:
Added OPEN PIPE for Windows. There still are problems receiving wide strings via pipe, but it works for executables with a wide string name and ANSI output.

Shouldn´t CURDIR return a path with a trailing (back)slash (which it currently doesn´t) just like EXEPATH does ?

jklwn4:
Currently FB converts wide string to ANSI for all file/path related functions. That is, you can only use wide strings representing ANSI strings, "real" unicode strings (e.g Russian characters) fail.

I replaced all ANSI entry points with wide string entry points in Windows. The conversion is now the other way round from ANSI to wide string in Windows if necessary.

For Linux wide strings are converted to UTF-8 where necessary. Travis (Linux) runs and all tests of the test suite run in Windows too.

Maybe there are still problems with cross compiling and this solution works only for Window and Linux, other targets are missing.

Ready for review and discussion.

JK

jklwn4:
The recent commits add the PR "File2" to this PR. Now (hopefully) all file related changes are unified in one single PR. This makes it possible to compile executables for beta testing.

jklwn4:
Jeff, Ralph,

why do i have to resolve a conflict in "file.bi", and why after resolving the conflict do i have to merge master into this branch in order to satisfy Github? And why doesn´t Travis run anymore?

PR #182, opened 2019-10-04
add new string processing functions

jklwn4:
This PR is not ready now. I will add a new string processing function with each commit.

JK


jklwn4:
It failed because of a problem Travis had. The code seems to be fine, except for a warning, which shouldn´t be: 

> src/compiler/pp-define.bas(610) warning 3(1): Passing different pointer types, at parameter 1 of HREPLACEW()

the corresponding code line is:

>DWstrConcatAssign( text, *hReplaceW( strptr(w), QUOTE, QUOTE + QUOTE ) )

where w is a wstring, so strptr(w) should be a wstring ptr (after recent changes), which is expected by HREPLACEW()

jklwn4:
Another Travis problem, now the first of 8 tests fails. Last time it was the fourth of 8.

XusinboyBekchanov:
Repeat copy is WString(ByVal count As Integer, ByVal ch_code As Long) As WString?

jklwn4:
REPEAT and (W)STRING are similar but not the same. (W)STRING repeats only the first character of a given string, REPEAT repeats the whole string. Example:
WSTRING(3, "asdf") -> "aaa"
REPEAT(3, "asdf")    -> "asdfasdfasdf"
...and REPEAT accepts all kinds of string arguments (z/w/u/string)

jklwn4:
Once again i changed the overload resolving logic for z/wstring, in order to avoid "ambigous ..." for z/wstring decisions, wstrings should be rated over zstrings. In case of ambiguity it´s safer to take the wide string string version of an overloaded function, even if this means to accept unnecessary conversions in some cases. 

Converting an ANSI string to wide string, processing it as wide string and converting it back is more time consuming, but doesn´t produce wrong results. Converting a wide string to ANSI, processing it as ANSI and converting back to wide string is bound to produce wrong results!

jklwn4:
This isn´t a serious pull request yet, i basically use it for testing, if it compiles in Linux. I wanted to port what i have as FB code to the RTL. Initally i tried FBSTRING, but couldn´t make it work. So i decided to make it step by step and use char for a start. 

Currently USTRING doesn´t support surrogate pairs, neither do statements like LEFT, RIGHT, MID. So we have support for UCS-2 and not UTF-16, which would be another task to make FB fully unicode compatible (in Windows).

rversteegen:
I wonder, what would be better, for all string functions like LEFT/RIGHT/MID to measure string lengths/positions in codepoints instead of wchars? I don't see any great solution. Probably most other languages measure in wchars?
Since UTF16 is a variable-width encoding I think it has negligible advantages over UTF8...

jklwn4:
In order to be able to process embedded nulls later on, i wrote my code from scratch. Currently i use strlen and wcstrlen for getting the lenght of a char or FB_WCHAR, which stop at the first null. Basically my code can process embedded nulls, if the correct lenght is given. I plan for passing this lenght for string/ustring later implementing a macro. The available C functions are written for char and wchar (null terminated strings), i fear there will be problems with embedded nulls. 

For UTF-16 LEN must return codepoints, but currently FB only counts wchars. Personally i don´t see any advantage of UTF-16 over UTF-8 too, but Windows doesn´t support UTF-8 natively, Windows internally uses UTF-16 - all ...A functions are merely wrappers for corresponding ...W functions. The good news is, surrogate pairs are rarely used in Western languages, it´s a requirement for Asian languages (and even MS still has problems here). So in almost any case UCS-2 (LEN = lenght in wchars, no surrogate pairs) is identical to UTF-16 and will do. Full support of UTF-16 would require substantial changes in FB. So for the moment i go for UCS-2 as does the rest of FB.

jklwn4:
Travis fails because of:

> chkdocs.bas(1043) error 18: Element not defined, fb_parse in 'wiki->Parse( sPages(j).sName, sBody )'

chkdocs.bas(1752) error 7: Expected ')', found ',' in 'wiki->Parse( "PrintToc", sBody )'

chkdocs.bas(1879) error 7: Expected ')', found ',' in 'wiki->Parse( sPages(i).sName, sBody )'

chkdocs.bas(2012) error 7: Expected ')', found ',' in 'wiki->Parse( sPages(i).sName, sBody )'

makefile:84: recipe for target 'chkdocs' failed

I don´t know what this means, the code compile fine and the tests run as well. A naming problem ?

jklwn4:
Same as before, what´s the problem with "fb_parse(string)", when all other additions are accepted?

jayrm:
`string.bi` defines `#macro parse()` which globally overrides every use of `parse` in source code, causing compilation failures where source already uses the symbol `parse`.  

Depending on usage, new names in the global namespace will potentially break user code.  Which is the main reason we avoid adding new symbols names to the global namespace.

I notice many of these new "keywords" are overloaded functions.  Could they be in an `#include` file rather than added by default by `rtl-string.bas`?

rversteegen:
With all these functions you're adding which have simple short names, naming conflicts are inevitable. However in this case wiki->Parse is a method so it would not clash with a function (builtin or not) called parse. The problem is that "parse" is a macro, not a function. I realise you did it this way to hide the declaration inside `string.bi`.

@jayrm: a lot of these functions have quirk syntax to support ANY args (like INSTR ... ANY "..."), so isn't declaring them in `rtl-string.bas` unavoidable? Well the alternative is to not use that syntax, and instead split out the ANY versions as separate functions.

If there was no quirk syntax, then you could declare these functions in string.bi like this:
```
extern "rtlib"
declare function parse overload alias "fb_StrParse" ( byref z as zstring, byval dummy as long, byval n as integer ) as string
declare function parse overload alias "fb_WstrParse" ( byref w as wstring, byval dummy as long, byval n as integer ) as wstring
declare function parse overload alias "fb_StrParseDelim" ( byref z as zstring, byval any as long, byref z as zstring, byval dummy as long, byval n as integer ) as string
declare function parse overload alias "fb_WstrParseDelim" ( byref w as wstring, byval any as long, byref w as wstring, byval dummy as long, byval n as integer ) as wstring
end extern
```

rversteegen:
Of course if the quirk syntax were actually implemented in the compiler, then I guess they wouldn't need to be declared in string.bi either. But that's a lot of work. It would be nice if `ANY` used as a prefix to an argument could be part of a function's call signature, so that you could overload based on its presence! Then the compile doesn't need to know anything about any of these commands, it would all just be normal declarations in `string.bi`.

I'm imagining something like
``declare function parse overload alias "fb_StrParseDelim" ( byref haystack as string, any byref needle as string, byval n as integer = 1 ) as string

jklwn4:
Ok, i understand the macro causes problems.

I added a new pp feature (#?) for implementing "ANY" (and maybe others) without the need for changing other things in the compiler. "#?" splits one parameter containg a space into two, the first one becomming a long (ASCII code of the first four characters of this word) and the following word as is. That is "..., ANY s, ..." becomes "..., &H00414E59, s, ..." it makes two parameters out of one and the additional first parameter can be parsed inside the receiving function. So i can have as many quirk words as i like without polluting the global namespace. The downside is, i need a macro for this.

In general i don´t quite understand the fear of adding new key words. A new feature needs a new key word, that´s inevitable i think. In C you have a very basic set of staments (key words) and you must add everything else by adding include files and libraries. BASIC (as i see it) offers a lot more than a basic command set and therefore must be allowed to have more reserved (key) words. 

The functions i want to add are extremly useful for string processing and therefore deserve being a part of the global namespace. I agree that not each and every addition must become part of the global namespace. I could put these macros into a new include file (maybe stringex.bi), but this would only shift the problem. Those wanting to include stringex.bi would have to deal with the naming conflicts anyway. 

So why not announce: there will be new features requiring new key words, these new reserve words will be ..., please adapt your code in order to avoid possible naming conflicts. The final decision is up to you Jeff.

jklwn4:
In CWiki.bi there is already a line (declare function Parse _), which uses "PARSE". As quick remedy i will rename my macro. This is exactly the situation i was speaking of in my previous post. "PARSE" seems to be the most obvious name, so in order to make it useable for FB´s syntax, the existing function would have to changed. 

I would do it, because then i have the most obvious name for FB. The alternative would be to have another (maybe worse, not so obvious) name for the new PARSE feature

Please don´t get me wrong, i´m describing my thoughts and preference here. I don´t claim that you should/must do it how i would like to have it - i´m discussing, i´m providing arguments.

jklwn4:
This is a Travis problem, 7 out of 8 pass, one fails.

jklwn4:
The following commits will place declarations of new RTL-function into "string.bi" and remove them from rtl-string.bas (which was used as a development step). All functions will be able to process embedded nulls for STRING and USTRING types.

jklwn4:
I replaced former ACODE/UCODE with COPY for copying raw data from a z/string to a w/ustring and vice versa. No conversion takes place, the data is just copied. This is necessary to avoid automatic  conversion.

Syntax: [z]string = Copy(wstring)
          w/ustring = Copy([z]string)

jklwn4:
Alas, it keeps failing at some test bescause of a declaration conflict, i don´t understand (it shouldn´t happen as far as i can tell)

jklwn4:
PATHNAME parses a path or file name to extract component parts and returns the requested part
These are the options for specifying the requested part:
PATH    Returns the path portion of the path/file Name. That is the text up to and
            including the last (back)slash (\/) or colon (:).

NAME    Returns the name portion of the path or file Name. That is the text to the right
              of the last (back)slash (\/) or colon (:), ending just before the last period (.).

EXTN    Returns the extension portion of the path or file name. That is the last
            period (.) in the string plus the text to the right of it.

NAMEX   Returns the name and the EXTN parts combined.

Syntax: resultstring = Pathname(PATH|NAME|NAMEX|EXTN, filespec)

jklwn4:
REPEAT returns a string consisting of multiple copies of the specified string. This function is a similar to STRING/WSTRING functions, but allows for strings of arbitrary length to be concatenated.

Syntax: resultstring = Repeat(5, "xyz")

jklwn4:
TALLY counts the number of occurrences of specified characters strings within a string. W is the string expression in which to count characters. M is a list of single characters to be searched for individually or in total. A match on any one of which  or a match in total will cause the count to be incremented for each occurrence. Note that repeated characters in m will not increase the count, if characters a searched individually. If m is not present in w, zero is returned. If "ANY" is not specified, m is handled "as string"

Syntax: n = tally(w <string to count in>, [any] m <string to find>)

jklwn4:
PARSECOUNT returns the count of delimited fields from a string expression. If w is empty (a null string) or contains no delimiter character(s), the string is considered to contain exactly one sub-field. In this case, ParseCount returns the value 1. m contains a string (one or more characters) that are seached individually or must be fully matched. If "ANY" is not specified, m is handled "as string". If m is not specified, it defaults to ",".

Syntax: n = parsecount(w <string to count in>, [any] m <separating string>)

jklwn4:
INVERT Inverts the contents of a string.

Syntax: resultstring = invert("xyz")

jklwn4:
INSERT inserts a string at a specified position within another string expression. Returns a string consisting of w with the string i inserted at position n. The first character in the string is position 1. If n is greater than the length of w then i is appended to w. If n is negative counting starts from right to left, -1 means the last character , -2 the last but one, and so on

Syntax: resultstring = Insert(w <string to insert in>, i <string to insert>, n)

jklwn4:
EXTRACT complement to the Remain function. Extracts characters from a string up to a character or group of characters. Returns a substring of w starting with its first character (or the character specified by nStart) and up to (but not including) the first occurrence of m. If m is not present in w (or is null) then all of w is returned from the nStart position. nStart is an optional starting position to begin searching and extracting. If nStart is not specified, position 1 will be used. If nStart is zero, a null string is returned. If nStart is negative, the starting position is counted from right to left: if -1, the search begins at the last character; if -2, the second to last, and so forth. If "ANY" is not specified, m is handled "as string"

Syntax: resultstring = Extract([nStart,] w <string to be searched>, [any] m <char(s) to be searched for>)

jklwn4:
REMAIN complement to the Extract function. Returns the portion of a string following the first occurrence of a character or group of characters. w is searched for the string specified in m, if found, all characters after m are returned. If m is not present in w (or is null) then a zero-length empty string is returned. nStart is an optional starting position to begin searching. If nStart is not specified, position 1 will be used. If nStart is zero, a null string is returned. If nStart is negative, the starting position is counted from right to left: if -1, the search begins at the last character; if -2, the second to last, and so forth. If "ANY" is not specified, m is handled "as string"

Syntax: resultstring = Remain([nStart,] w <string to be searched>, [any] m <char(s) to be searched for>)

jklwn4:
PARSE returns a delimited field from a string expression. m contains a string of one or more characters that must be individually or fully matched to be successful dependig on "Any". If n evaluates to zero or is outside of the actual field count, an empty string is returned. If n is negative then fields are searched from the right to left in w. M is case-sensitive. If m is not specified, it defaults to ","

Syntax: resultstring = parse(w <string to parse>, [[any] m <delimiter string>,] n <position>)

jklwn4:
SHRINK shrinks a string to be able to use a consistent single character delimiter. The purpose of this function is to create a string with consecutive data items (words) separated by a consistent single character. This makes it very straightforward to parse the results as needed.
If m is not defined then all leading spaces and trailing spaces are removed entirely. All occurrences of two or more spaces are changed to a single space. Therefore, the new string returned consists of zero or more words, each separated by a single space character. If m is specified, it defines one or more delimiter characters to shrink. All leading and trailing mask characters are removed entirely. All occurrences of one or more mask characters are replaced with the first character of wszMask The new string returned consists of zero or more words, each separated by the character found in the first position of m. WhiteSpace is generally defined as the four common non-printing characters: Space, Tab, Carriage-Return, and Line-Feed. m = (W)Chr(32,9,13,10)

Syntax: resultstring = Shrink(w <string to shrink>, m <char to stay + chars to remove>)

jklwn4:
REMOVE(ME) returns a copy of string w with substrings m removed individually or in total. If m is not present in w, all of w is returned. If "ANY" is not specified, m is handled "as string"

Syntax: resultstring = Remove(me)(w <string to remove from>, [any] m <string to remove>)

jklwn4:
REPLACE within w replace all occurrences of one string with another string or all occurrences of any of the individual characters specified in the m string with r. The replacement can cause w to grow or shrink in size. When a match is found, the scan for the next match begins at the position immediately following the prior match. r can be a single character or a word. If "ANY" is not specified, m is handled "as string"

Syntax: resultstring = Replace(w <string to replace in>, [any] m <char(s) to be replaced>, r <replacement string>)

jklwn4:
It passes all my own tests, it compiles in Windows and Linux - ready for beta testing.

Maybe a SPLIT function is missing (split an existing string into delimited parts and copy these parts into an array, dimension this array accordingly)

After successful testing as a last step the A and W functions should go into separate files named according to the current naming convention. Currently it´s still convenient to have everything in place.

PR #183, opened 2019-10-05
update sqlite 3 headers to v3.30.0 + fix example

swurzinger:
Updated SQLite headers to v.3.30.0 using the headers from https://www.sqlite.org/2019/sqlite-amalgamation-3300000.zip

Fixed compiler warning and possible 64-bit compatibility issue in the SQLite example program.


jayrm:
Updated wiki documentation on [SQLite](https://www.freebasic.net/wiki/wikka.php?wakka=ExtLibsqlite)

PR #184, opened 2019-10-05
update BASS 2.4 headers

swurzinger:
Updated headers from current BASS 2.4.14 headers 
http://us.un4seen.com/files/bass24.zip


jayrm:
github has an interface for resolving conflicts.  And in this case, it would just be changelog.txt.  Resolving the conflict this way makes a correct but overly complicated merge history, especially since it is obvious from the pull request that other work is not going to get clobbered.  I guess it just boils down to personal preference that I would rather rebase and force push on top of current master, in these kinds of cases.

Updated wiki documentation and example on [BASS](https://www.freebasic.net/wiki/wikka.php?wakka=ExtLibbass)

PR #186, opened 2019-10-06
update curl headers to 7.66.0

swurzinger:
updated curl binding to curl v7.66.0 from https://curl.haxx.se/download/curl-7.66.0.tar.gz
updated changelog too this time ;-)


jayrm:
Works for me when build `fbdoc` manual generator.

PR #189, opened 2019-11-02
fix error with LEN and explicit namespace prefix

jklwn4:
This fixes an error with LEN and explicit namespace syntax (namespace.identifier). before it was necessary to enclose such syntax in parenthesis to make it an expression, now it works as expected (without extra parenthesis)

Jeff, i hope this is correct!

JK

### DIFF
--- a/src/compiler/parser-decl-symbtype.bas
+++ b/src/compiler/parser-decl-symbtype.bas
@@ -316,12 +316,21 @@ function cTypeOrExpression _
 
 	if( maybe_type ) then
 		'' Parse as type
-		if( cSymbolType( dtype, subtype, lgt, is_fixlenstr, FB_SYMBTYPEOPT_NONE ) ) then
+
+        dim as LEX_CTX lex_save
+        lex_save = lex                                '' save lexer status
+
+        '' this might "eat" namespace prefixes, which leads to a "Varaible not declared" error with explicit 
+        '' namespace prefix syntax (namespace.identifier)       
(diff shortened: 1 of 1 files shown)


jklwn4:
As far as i could find out, the problem here (src/compiler/parser-decl-symbtype.bas: cTypeOrExpression, line 319) seems to be that, cSymbolType eats the namespace prefix. So afterwards, in case it wasn´t a type, the lexer has advanced one position - it isn´t where it was before. I simply restore the lexer context to what it was, before cSymbolType is called. According to my tests this seems to fix the issue and i didn´t discover any unwanted side effects so far.

jayrm:
The design of the compiler is that it is always looking forward, translating the code in one pass.  Very rarely, or only in very specific cases do we restore previous lexer/parser/ast/ir/emit states, and even then only under very specific conditions.  In other words, once we get information from the source code, we either use it right away, or save it for later.

`LEX_CTX` can't be saved / restored using the method in this PR.  The lexer context is not fully restored using this method because of the use of file buffers and other dynamic objects.  Besides, the LEX_CTX structure is like 400KB+, so not the kind of thing we would save a copy of.

Solving the issue as a lexer problem is very difficult as would need to change a lot of the lexer design to restore a previous state.  Handling the issue as a parser problem makes it much easier to deal with.

Proposed fix for sf.net # 404 is in PR #201

jklwn4:
So analysis was correct, solution wasn´t. Thanks for the insight. I´m going to close this PR now.

PR #190, opened 2019-11-05
Direct2D Windows gfxdriver

adeyblue:
Perhaps slightly less desirable now the DirectDraw one was fixed, but still.


jayrm:
Hi @adeyblue,

First impression is that this looks OK to me.  I'm no expert in this area so it would help to know what I am looking at.  

1) In what cases would this driver be preferred over the existing driver?  
2) Major benefits: either code itself or runtime? or any limitations?
3) Any possibility you know of where we may need to exclude this driver?  Either at compile time or run-time if building of running on older system?

adeyblue:
1 & 2) Well like I said, I started this before the DirectDraw one was fixed so the main benefit at that point was that it actually worked on 64-bit. And before I fixed the GDI driver in the other PR, this handled all the screen modes too. 

What this is is basically a superset of the DirectDraw driver:
That one doesn't handle GFX_SHAPED_WINDOWs so anybody wanting that falls back to the GDI one. This one does it natively.

The current DirectDraw one also doesn't VSync properly in some situations, this one does

This one also reacts to input events and messages when they occur rather than polling for them like the other two, so it's slightly more efficient for that too.

In those senses it's technically better than the current DirectDraw driver - like all these GFX drivers though, it's still just a glorified memcpy from the framebuffer to the screen. Direct2D is Microsoft's current bees knees, like DirectDraw was back in the day. That driver was probably written because DirectDraw used to be the path to hardware acceleratation. Now Direct2D is the hardware accelerated 2D API. So this driver is also 'better' in that sense too.

3) This driver trivially fails at runtime if the OS is Vista or earlier (where Direct2d doesn't exist) or if D2d1.dll can't be loaded, making gfxlib go try the others. The only hard dependencies this has are those functions that are in NT3.1, with anything newer dynamically loaded. PR180 with the wide string RTL does more runtime requirement 'damage' than this, since that (currently, avoidably) bumps up the baseline OS requirement to NT4, despite this being a much newer technology. 

In its current state, if FB is built with either DEBUG or D2D_DEBUG defined, this driver turns on the Direct2D and 3D debug runtimes. Those require extra installs from Microsoft's SDK or the driver will failover back to the DirectDraw one. Which things are required and from where are detailed in the pastebin link in the source comments (I wrote the pastebin btw). No additional installs are required to use the driver in non-DEBUG builds. 

At build time, the DirectDraw one is excluded from cygwin builds for some reason, I just copied that ifdef since I guessed if cygwin didn't have headers for that it definitely wouldn't have headers for this. Building this doesn't require anything other than an environment that contains the Direct2D headers. The current MinGW packages the FB Wiki points people to in order to build FB is enough to compile it, since those are the instructions I used to build this. I've no idea about any cross-compilation shenanigans.

TL;DR This is the graphics equivalent of the wide string API in PR180. You don't need it for FB to work, it doesn't get you anything you didn't already have, it's just slightly more efficient and optimized because it's the current Microsoft way to do things.

jayrm:
Perfect, thanks for the explanations.  I will rebase this to master and add a changelog message.

PR #191, opened 2019-11-08
Macro and preprocessor related changes

jklwn4:
This PR is ready now

It summarizes macro and preprocessor changes from other PRs allowing for a more flexible macro syntax.


jklwn4:
No checks by Travis anymore - why?

jklwn4:
#? :
new preprocessor operator for implementing quirk words like "any". This operator makes two comma separated parameters out of one parameter. If the original parameter contains a space (outside of a quoted string literal), the parameter is split into two parts. The first part is converted into a number, the ascii codes of the first four characters of the the first part form a LONG. The second part is passed as is.
If there is no space, "0, <original parameter>" is returned. In effect leading quirk words are separated from the actual parameter and passed as an extra parameter in form of the Ascii-code of the leading four characters of the quirk word. This allows for different quirk words. Passing it as string, would bias overload resolution. Converting and passing as number, doesn´t raise problems in this regard.

jklwn4:
#REDEF :
allows for temporarily re-defining (and thus overriding) an already existing #define (which was forbidden before). This new define is valid only in the current scope and is automatically restored to it´s previous definition when a scope ends. In global scope you may use #UNDEF to restore the previous #define. This allows for otherwise forbidden key words to be used in macros and elsewhere

jklwn4:
Macro parameter check modification:
the compiler checks the parameter count of macros and throws a compile error, if the actual number of parameters doesn´t match the definition. This check has been disabled in case of a variadic macro parameter. That is, a variadic parameter may now not only be missing at all, but also qualify as one or more parameters. A variadic parameter may have commas inside, which in effect makes it more than one parameter. A macro may now also receive less parameters than specified in it´s definiton.

jklwn4:
"###" :
new preprocessor operator, which returns the number of arguments (count of commas + 1) in a variadic macro parameter.

"#&#" :
stringize uppercase, that is a macro parameter is returned as uppercase string - same as "#", but uppercase.

jayrm:
@jklwn4  
> No checks by Travis anymore - why?

Not sure why.  The builds are still being triggered but the result is not being reported in github.

See: https://travis-ci.org/freebasic/fbc/pull_requests

Just a suggestion: should probably get in the habit of running the test suite on your own local PC before creating the pull request.  Optionally, you could also create your own Travis CI account and link your github account to it.  You can then see the Travis CI results on your own github branches.

I had to play around with my github & Travis CI account settings to get it working again (I hope) for my own accounts & branches.

jklwn4:
"#%" :     
concatenate space separated parts of an argument, return the argument with spaces removed.

jklwn4:
Thanks Jeff!

jklwn4:
Nice, Tavis runs again and i know at least which new feature is still buggy.

jklwn4:
"#\#" makes foo, (bar) out of foo(bar). That is, it separates function name from arguments. If the first character of the argument is a bracket, the argument is returned as is plus an empty argument. e.g
"(abc)" -> "(abc),"

jklwn4:
closed in favour of a new one

Issue #192, opened 2019-11-10
enable framebuffer driver for arm


stephaneweg:
in the raspberry pi, we want to use the framebuffer driver, to enable graphic mode without x11
the header files disables this drivers for the arm, because there is inline assembly.

this inline assembly is only used for the 4 bit per pixel mode, but we will use the 16bpp mode, so we could put the content of this method in comment for the arm target, instead of disable the driver

jayrm:
framebuffer enabled and 4-bits-per-pixel blitter added in #256 
Closing issue as resolved.

PR #193, opened 2019-11-13
Disable vga16 blitter on ARM because it uses x86 asm (as opposed to disabling the whole fbdev driver)

countingpine:
Drive-by patch for Issue #192.
Completely untested.


jayrm:
Patch from this PR folded in with #256 - for history only as the changes are overwritten by follow-up commits.

PR #194, opened 2019-11-13
GDI Gfxlib Driver Scanline Size Fix

adeyblue:
Graphic screens 2 & 8 only render to half the window using the GDI driver since it doesn't take into account the scanline size.  This PR fixes that by using a blitter, and cleans up some other stuff:
a) only allocating the blitter buffer when necessary
b) fixing exit from full screen if window creation fails (the change back depended on the window being created)
c) alpha_remover_blitter was broken in the >1 scanline case it attempted to handle (gfxlib only writes upto __fb_gfx->h, not __fb_gfx->h*scanline_size) - but as alpha_remover_blitter is only used in 32bpp modes, and there are no 32bpp modes with a greater than 1 scanline size, I removed the multiplication and stuck the conditions in an assert.

Example code that this PR now renders correctly:
Screen 2
https://www.freebasic.net/forum/viewtopic.php?t=19967#p175270

Screen 8:
https://www.freebasic.net/forum/viewtopic.php?t=21047#p185975

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -40,6 +40,7 @@ Version 1.08.0
 - github #205: fix rtlib potential malloc null ptr access in dev file read/write methods
 - default new, new[] operator (allocate) after #undef allocate caused compiler crash
 - default delete, delete[] operator (deallocate) after #undef deallocate caused compiler crash
+- windows GDI gfxlib driver now displays double scanline screen modes correctly (screen 2 & 8), rather than half screen (adeyblue)
 
 
 Version 1.07.0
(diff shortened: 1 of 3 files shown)


jayrm:
I tested the samples code linked, before and after and appears to work as needed.  I have rebased on to current fbc/master.

Only change I made was to the order of driver list to now be "DirectX", "GDI", "Direct2D".  In the new Direct2D driver (PR #190) the updates to screen only appear to be done if there is user input, i.e. `MsgWaitForMultipleObjects(... QS_ALLINPUT)`, or if the fb gfx internal buffer is dirty.  If the window is in the background (no input), and screen is not being updated (for example the program is sitting in a `SLEEP` statement) then screen does not get updated.  Moving the window off screen, or overlapping a foreground window causes areas to erase.  I moved the Direct2D driver down in the selection list so that it is still available to users to try out, but not yet the default.  To specifically select can use `screencontrol fb.SET_DRIVER_NAME, "DIRECT2D"`

PR #196, opened 2019-11-25
Update fb_gfx_linux.h

stephaneweg:
fbdev can be enabled on arm platform. the only issue is that the vga16 blitter must be disabled in this case. So i propose to remove the macro  #define DISABLE_FBDEV when we are not on a X86 platform
instead of this, ce could add a directive in the vga16_blitter code


jayrm:
Concepts from this PR folded in with #256 - thank-you

PR #197, opened 2019-11-25
Update gfx_driver_fbdev.c

stephaneweg:
fbdev can be enabled on arm platform. the only issue is that the vga16 blitter must be disabled in this case. So i propose to remove the macro  #define DISABLE_FBDEV when we are not on a X86 platform
instead of this, ce could add a directive in the vga16_blitter code 

Note: i have tested this on my raspberry pi, and it work fine without X11 with only the framebuffer

### DIFF
--- a/src/gfxlib2/linux/gfx_driver_fbdev.c
+++ b/src/gfxlib2/linux/gfx_driver_fbdev.c
@@ -17,7 +17,11 @@
 #define FB_AUX_VGA_PLANES_VGA4	0
 #endif
 
+/* Disable fbdev driver on ARM, because it currently uses x86 inline asm */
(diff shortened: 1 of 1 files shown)


countingpine:
Just to say that this, along with #196, comprises an alternative fix I did in Pull Request #193 for Issue #192.  (Tagging them all to help interconnect them.)

I decided it was better not to have the function exist at all outside x86, and so changed the function that references it to return -1 instead, which is what happens in a different case if blitter is NULL.

My changes haven't been tested, but I'd welcome feedback if you are able to test.

jayrm:
Concepts from this PR folded in with #256 - thank-you

PR #205, opened 2019-12-12
Leaky realloc fixes

adeyblue:
There are also eight other leaky reallocs in the unicode string conversion like this one:
https://github.com/freebasic/fbc/blob/5e93d69f151172694f98f08376838e90f632d6b9/src/rtlib/utf_convto_char.c#L55
As none of them are checked for failure right now it's above my pay grade to decide what should happen in the failure cases


jayrm:
@adeyblue, thank-you.

The string conversion functions (utf_conv*.c) appear to have consistent pattern of intended use.  If the passed in dst string is null then the function should malloc/realloc a buffer for the converted string and caller is responsible to free.  If the passed in dst string is non-null, the function should never realloc it.  I think is safe to return null pointer on failed malloc/realloc and I have updated the functions.  I added some code comments to help explain the usage of these functions.  I'll leave the other unchecked malloc/calloc calls found elsewhere in the rtlib for future effort.

PR #206, opened 2019-12-13
expand PARSE statement, PARSE ANY, dst ptr, src ptr, count

jklwn4:
This expands the PARSE statement. "PARSE ANY, destination ptr, source ptr, count" copies count bytes from source memory to destination memory. If memory buffers overlap, source gets overwritten in the overlapping area. In case of null pointers or a count of zero, the statement fails and error "illegal function call is set". Otherwise you have to make sure you supply pointers to (enough) valid memory, or you will cause a GPF.


jayrm:
I assume title and description should be 'POKE ANY...'

I agree with the concept.  It would be very useful to have a builtin low level way to move/copy memory directly rather than writing a user function or including all of ./inc/crt/mem.bi.  And the `POKE` keyword is already a reserved quirk keyword, so expanding on the syntax to `POKE ANY, dst, src, count` is reasonable.  It performs a `POKE` relateable function and does not create any conflicts.  We have a number of low level memory operations supported directly by the compiler ( [c|re|d]allocate, clear as aliases for the c runtime malloc, calloc, realloc, free, memset ).  These functions are both required by the compiler and are used in many normal freebasic programs.  fb_MemCopy used to be defined in older versions of fbc.  The existing `fb_MemCopyClear` is from earlier days and uses signed buffer lengths.  in 2013, dkl fixed the low level memory functions to correctly accept unsigned length values, so should not be used as an example for adding low level memmove.  Assuming memmove is available on all our targets, then there is no real reason to wrap it in a fbrtlib function.  Better example is the `clear` function alias for `memset` which is declared as `byref as any` on the fbc side and `unsigned char *` on the fb rtlib (crt) side for referring to the dst argument/parameter.  Only if there is a platform that does not have a native memmove function in the crt, then a wrapper would need to be added to fb runtime.

There are a couple of issues with the implementation of cPokeAny in this PR
- `POKE ANY PTR, dst, src` is valid syntax for pointer types so need to take care directly lexing the 'ANY' token
- if `astNewArg` fails (returns NULL) then need to clean up the invalid AST objects before returning and implement the Error handling/recovery code.
- need some tests that prove usage

Just for fun, I have rebased this PR on current master and created small commits to show the incremental changes to get to a final implementation.

jklwn4:
Thanks for taking care of this! 
As always i try to learn from it. Looking at your changes it seems my code wasn´t that bad.  Apart from a lot of formal things (place, name, etc.) and apart from the signed/unsigned bug it basically worked. Thanks again.

JK

PR #210, opened 2020-01-10
GfxLib: Update dirty bits in WM_PAINT

adeyblue:
This makes WM_PAINT signal which lines needs redrawing in response to the message by setting the relevant fb_gfx->dirty bits. Handlers of some other messages signal redraws by setting these bits, why not the one that actually signals redrawing?

This is a fix for the comment about the D2D driver not redrawing after being uncovered in #194 , which wasn't because of dodgy MsgWaitForMultipleObjects flags, but because the dirty bits weren't set in response to the WM_PAINT invalidation and the D2D driver uses the dirty bits as a signal of what to transfer from the framebuffer to the display - cos that's what they're for, right?

Also fixed the function pointer warning because I was tired of seeing it and it wasn't noise, the prototype was incorrect when compiling for 64-bit. The timer ID is a 64-bit value on 64-bit, not 32.

### DIFF
--- a/src/gfxlib2/win32/gfx_win32.c
+++ b/src/gfxlib2/win32/gfx_win32.c
@@ -120,7 +120,16 @@ static void ToggleFullScreen( EVENT *e )
 	has_focus = FALSE;
 }
 
-static VOID CALLBACK fb_hTrackMouseTimerProc(HWND hWnd, UINT uMsg, UINT idEvent, DWORD dwTime)
+static VOID CALLBACK fb_hTrackMouseTimerProc(
+	HWND hWnd, 
+	UINT uMsg,
+#if _WIN64 
+	UINT_PTR idEvent, 
+#else
+	UINT idEvent,
+#endif
+	DWORD dwTime
+)
 {
 	POINT pt, rect_pt[2];
 	RECT *rect = (RECT *)rect_pt;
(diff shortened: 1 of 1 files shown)


jayrm:
Ah yes, seems I misread docs for MsgWaitForMultipleObjects and wrote a bad test.  Can see the intent there now.  Thanks.  My overall impression of the Direct2D driver so far that it is cleaner and easier to follow than the existing drivers.  However, it's still a little confusing to me where the mutex locks need to be, and if we actually have them in all the needed places.  Global data (like dirty bits) could be updated from the user thread or the driver thread.

Setting dirty bits in response to WM_PAINT seems like right thing to do.  The dirty bits are usually only ever set (I think) when we are drawing to active screen and not in between a screenlock/screenunlock.  It raises the question for me though, why we don't see this already in existing gfx drivers?  Maybe for more precise timing? or to follow a refresh rate? I don't know.

By merging this change in, I have a feeling will have to look at cleaning up the existing drivers.
1) It's weird to me that gdi_paint() calls InvalidateRect() in between BeginPaint() and EndPaint()
2) In the existing drivers, most of the time we don't really care what happens in WM_PAINT because we are regularly painting the screen from the thread.  Really just validating the rect, I think.
3) Not sure what will happen on old versions of windows.
4) Maybe we want to have some variation of the "manual" versus "automatic" refresh idea that was added to OpenGL driver?  i.e. additional screencontrol bits to modify how the driver might handle screen paints and or timing of updates.

Anyway, I think we can keep moving forward with this, and I will followup with some news in [Need new gfxlib driver for Windows 7, 8, 10](https://www.freebasic.net/forum/viewtopic.php?f=6&t=26799&p=267876#p267876)

PR #211, opened 2020-01-22
Macro and preprocessor related changes (2)

jklwn4:
This PR is ready now.

Summarizes macro and preprocessor changes from other PRs allowing for a more flexible macro syntax.

- TYPEOF now works in regular code as well
- remove error for failing parameter count check when there is a variadic macro parameter
- allow for macros without enclosing brackets for parameters under certain conditions


jklwn4:
Introduced a new syntax, besides the already existing pp operators (#, ##) all new operators work like this: #<key word>[(literal string)]#

e.g.: #ucase#, #left_of(xyz)#

the latter would return everything to the left of "xyz" (case insensitive) of a parameter. This syntax is expandable to all kinds of possibly useful string (parameter) processing functions.

The first one here is "stringize uppercase" (#ucase#), which makes case insensitive comparisons (BASIC is a case insensitive language) inside the preprocessor possible. 

A corresponding #lcase# could be added, but i don´t know what for, because case insensitive comparisons can already be done with #ucase#. 

Travis still works, but the result isn´t reported here.

jklwn4:
Sorry, pushed the wrong button! :-(

jklwn4:
#count#: count number of arguments in a vriadic parameter by counting commas. The returned number is the number of commas + 1. Even an empty paramter counts as 1.

#join#: concatenate spaces separated parts of a parameter, i.e. remove all spaces

#split#: insert a comma between space(s) separated parts of a parameter. Preserve space(s) in literal strings and in expressions enclosed in brackets. This may change the number of parameters. In case there are no spaces, the original paramter is returned.

#left_of(<token>)#: return everything to the left of <token>, where <token> must be a literal string (without enclosing quotes). In other words, cut before <token>.

#rigth_of(<token>)#: return everything after  <token>, where <token> must be a literal string (without enclosing quotes). In other words, cut after <token>.  

#is(<token>#: return 1, if parameter = <token>, , where <token> must be a literal string (without enclosing quotes). The original parameter is returned otherwise. 

All comparisons (<token>) are case insensitive, because BASIC is a case insensitive language

PR #239, opened 2020-09-17
FB_READ_LARGE / FB_WRITE_LARGE functions.

countingpine:
They differ from fread/fwrite in two main ways:
1. They take only a single parameter for the size, in bytes
2. They read/write in chunks of 1MB, to bypass any multi-gigabyte limits.
1MB chunks should be large enough to avoid significant overhead per chunk,
yet well short of any 32-bit length limits.

Use them in place of fread/fwrite in fbDevFileRead/Write, called eventually
by file GETs/PUTs.

Notes:
I'm not necessarily happy with the location of the functions, they may be better in separate modules, or at least somewhere that changes don't trigger a full rebuild.  I think the main thing at this point is just to submit something that works and let it be refined, otherwise the code is likely to rot on my hard disk.

It may also technically be worth using them inside Bload/Bsave functions, although 2GB images would be pretty rare.  There shouldn't be much overhead on small reads/writes though.

### DIFF
--- a/src/rtlib/fb_file.h
+++ b/src/rtlib/fb_file.h
@@ -365,3 +365,68 @@ extern const UTF_8 __fb_utf8_bmarkTb[7];
 
 FBCALL int          fb_FileCopy         ( const char *source, const char *destination );
 FBCALL int          fb_CrtFileCopy      ( const char *source, const char *destination );
+
+
+
+#define FREAD_CHUNK_SIZE 1048576
+#define FWRITE_CHUNK_SIZE 1048576
+
+static __inline__ size_t FB_FREAD_LARGE( void *ptr, size_t nmemb, FILE *stream )
+{
+   size_t total = 0, nread;
+
+   /* if nmemb <1MB, use fread() directly */
+   if (nmemb < FREAD_CHUNK_SIZE) return fread( ptr, 1, nmemb, stream );
+
+   while (nmemb > FREAD_CHUNK_SIZE) {
+      /* read chunk */
+      nread = fread( ptr, 1, FREAD_CHUNK_SIZE, stream );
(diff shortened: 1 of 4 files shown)


src/rtlib/fb_file.h
```diff
@@ -365,3 +365,68 @@ extern const UTF_8 __fb_utf8_bmarkTb[7];
...
+
+#define FREAD_CHUNK_SIZE 1048576
+#define FWRITE_CHUNK_SIZE 1048576
+
+static __inline__ size_t FB_FREAD_LARGE( void *ptr, size_t nmemb, FILE *stream )
+{
+   size_t total = 0, nread;
+
+   /* if nmemb <1MB, use fread() directly */
+   if (nmemb < FREAD_CHUNK_SIZE) return fread( ptr, 1, nmemb, stream );
```
rversteegen: These shortcut 'if's are pretty pointless, only making the functions longer.

countingpine: Yeah, you're probably right. When I originally wrote the function I kept separate size and nmemb parameters, and it was useful to bypass the more complex logic.
Now though, it just saves potentially a tiny amount of arithmetic.

jayrm:
Maybe can use in `src/rtlib/file_copy_crt.c` too.  Perhaps with an optional parameter on fbc's `FILECOPY` for buffersize,
I think `fb_file.h` is OK place if do intend to use inline and I wouldn't worry about the full rebuild.

I have an other PR to do, first then I will rebase this to new master and merge it in _(rebase is my preferred strategy)_.

jayrm:
Some weirdness with branches and travis-ci.  I think we all good though.

PR #240, opened 2020-09-20
Check for fixed length arrays when using ERASE & REDIM

jayrm:
Use new flags in the array descriptor (as of fbc-1.08) to make run time checks on the type of array: either static (fixed length) or dynamic (variable length)

When an array is passed as a parameter to a procedure, it is unknown at compile time if the array will be static or dynamic.  Previously, all arrays passed to procedures were handled as though dynamic, causing arrays to be destroyed or changed when they should not be.

With this change:
`ERASE` will perform run-time check to determine if the array is static.  If it is, array will be cleared only
`REDIM [PRESERVE]` will throw a run time error if the array passed to the procedure is static

Several internal changes and a few API changes were required to make this fix.  Because of this opportunity was taken to remove some of the legacy bits of the rtlib as it relates to arrays.


rversteegen:
Cause for celebration! Thank you. That lack of runtime type checking has been really bothering me for 15 years.

PR #244, opened 2020-09-21
Fixes sourceforge bug #918

adeyblue:
Since GCC doesn't support SEH to catch failing allocas and with malloca being relatively new invention, this seems the path of least resistance.


jayrm:
Yes, @adeyblue .  Looks good to me.  Safe and stable.  No doubt the alloca was intended to be 'faster', but if it sometimes fails... no so great.

FYI, rebase and added changelog.txt message coming.

Issue #216, opened 2020-06-16
function getMouse() seems broken on Win10 when using the Console Mode (screen 0)


ghost:
running the sample code from the manual 1.0.7 about getMouse()
the return of x, y, buttons works when setting any screenres but it doesnt return anything when in a console window. it seems to have worked on win7... something must have changed for win10

jayrm:
Confirmed that behaviour is different between win 7 & win 10 and that #245 appears to correctly handle both versions.

jayrm:
fixed by #245

PR #245, opened 2020-09-22
Turn off the QuickEdit console mode during GetMouse

adeyblue:
This fixes what is likely occuring with #216. If QuickEdit is enabled, no mouse messages are generated so GetMouse data won't ever change.

I cleaned up the repeated initialisation and removed the #if-s I think after 8 years of being there the current code has been found to work.


jayrm:
Tested on win7 & win10.  rebase and updated changelog.txt coming.

Issue #246, opened 2020-09-27
rtlib leaks thread local data


adeyblue:
The thread local data stored in rtlib can be left allocated in certain situations. That stored by Dir is the most easily noticable. Open Task Manager to the Details tab (Win10) or Processes tab (<=Win 7), turn on the Handles column, and run the code below. The count of handles will rise as the reported thread count does (well, as long as %ProgramFiles% has at least one directory in it).

It happens as Dir only closes the directory enumeration handle when everything has been enumerated. If you stop before the end, it is never closed. The same thing looks to happen with fds on Linux (I can't test on there, but the Linux Dir code is patterned the same way as the Windows one).

Memory can also leaked by the Input statement which stashes the typed input in a thread local structure (src/rtlib/con_input.c). 

    #include "crt/stdlib.bi"
    #include "dir.bi"
    
    Dim Shared g_threadCnt as Long = 0
    Dim Shared g_countMutex as Any Ptr
    Dim Shared g_stop As Long = 0
    
    #ifdef __FB_WIN32__
    #include "windows.bi"
    
    Const ENV_VALUE As String = "ProgramFiles"
    
    #else
    
    Const ENV_VALUE As String = "HOME"
    
    #endif
    
    Private Function IncrementLong(ByRef count As Long) As Long
    
        MutexLock(g_countMutex)
        count += 1
        Function = count
        MutexUnlock(g_countMutex)
    
    End Function
    
    Private Sub ThreadProc(ByVal p As Any Ptr)
    
        Dim threadId As Long = IncrementLong(g_threadCnt)
        Dim pDirpath As String Ptr = p
    
        Dim folder As String = Dir(*pDirPath, fbDirectory)
    
        If Len(folder) = 0 Then
           Print "Dir failed in thread id " & Str(threadId)
           IncrementLong(g_stop)
        End if
    
    
    End Sub
    
    #define THREADS_PER_LOOP 32
    
    Private Sub Setup()
    
        Dim zDirPath As ZString Ptr = getenv(ENV_VALUE)
        If zDirPath = Null Then
            Print "Couldn't find env var " & ENV_VALUE
            Exit Sub
        End If
    
        Dim dirSpec As String = *zDirPath & "/*"
        Dim threads(0 to (THREADS_PER_LOOP - 1)) As Any Ptr
        Dim i As Long = 0
        Dim created As Long = 0
        Dim loopCount As Long = 0
    
        Print "DirSpec = " & dirSpec
    
[...]

PR #247, opened 2020-09-27
TLS Data Destructors

adeyblue:
This is a fix for #246. Somebody already noticed thread local data leaking was a problem with the special case freeing of the GFX thread data in src/rtlib/thread_core.c, but the problem extends to Dir and Input which may also have things to cleanup.

This PR generalises the special case so that a destructor function is passed when the thread local data is requested/allocated. Now when the thread stops, any destructors present are called to release the outstanding resources.

It's not tested in *nix or Dos


jayrm:
Excellent write up in #246 
Pull request looks good.  I was able to test in Linux and DOS.  Some minor typo's otherwise should be good to go.
I will update the DOS stuff separately as I forgot to include in my commits to this PR.

- tests in DOS and Linux
- changelog.txt updated

PR #252, opened 2020-10-12
Build 1.08 updates

jayrm:
Update of some build options for building on WinXP and win32-mingworg and DOS/DJ.

New DX10 driver for gfxlib2 disabled for win32-mingworg builds.  If `DISABLE_D3D10` is defined when compiling gfxlib2, driver is excluded.  When building on WinXP, try to detect winnt 5.1 and automatically define DISABLE_D3D10.

Fix failing `crt/argv.bas` test on DOS.  When building dos version on WinXP w/DJ, short versus long filenames can cause the test to fail.  Test is disabled with an `ENABLE_CHECK_BUGS` conditional in the test file.

In the `contrib/release/build.sh` release script, the FBSHA1 value was getting lost over separate build steps.  Also, fbc's `makefile` tries to use `git` to determine the FBSHA1 value if it is not set.  This can fail during automated rebuilds when the PATH is modified and `git` no longer available.  Instead, FBSHA1 captured during the checkout from repository and saved for later use by the script.

Update for `SYMB_REG.TXT` on DOS for the rtlib exports (used by the DLL code).

Plus over last days, have tested full release generation plus test-suite on:
- Ubuntu 14.04, 16.04, 18.04, x86
- Ubuntu 14.04, 16.04, 18.04, x86_64, gcc & gas64 backends
- WinXP (win32-mingworg & DOS)
- mingw-w64/mingw32 under win7
- mingw-w64/mingw64 under win7, gcc & gas64 backends
- FYI, my build environment is completely broken on Win98SE.  It was painful on the last release, so I have no plans to fix.

Remaining work (for another time ... not now):
- A new LibFFI 3.3 was released since our 1.07.1 release, so LibFFI should get updated (so we are building our own version again if possible).
- `build.sh` for DOS is still set for old versions of gnu tools so the download paths are incorrect.  Tests were made with cached versions of the downloads, and new versions need to be selected and tested.
- `build.sh` still using gcc 5.2.0 still as default for mingw-w64 and needs to be updated to most recent available and tested.

### DIFF
--- a/contrib/release/build.sh
+++ b/contrib/release/build.sh
@@ -187,6 +187,10 @@ if [ ! -z "$repo_url" ]; then
 	git reset --hard "$fbccommit"
 fi
 
+# grab the FBSHA1 now, instead of forcing fbc/makefile to do it
+# and we don't want to rely on host having git it the path
+FBSHA1="$(git rev-parse HEAD)"
+
 cd ../..
 
 cd build
@@ -458,7 +462,7 @@ set PATH=$dospath/bin
 
 echo bootstrapping normal fbc:
 cd fbc
-make FBC=../$bootfb_title/fbc.exe
+make FBSHA1=$FBSHA1 FBC=../$bootfb_title/fbc.exe
 if ERRORLEVEL 1 exit /b
 make install
 if ERRORLEVEL 1 exit /b
@@ -468,7 +472,7 @@ echo rebuilding normal fbc:
 cd fbc
 make clean-compiler
 if ERRORLEVEL 1 exit /b
-make FBSHA1=1
+make FBSHA1=$FBSHA1
 if ERRORLEVEL 1 exit /b
 make install
 if ERRORLEVEL 1 exit /b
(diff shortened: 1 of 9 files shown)


jayrm:
Sometimes, there will be a failure in a job due to the test-suite, but passes on re-running the job.

I see now that it is due to testing multiple threads but the fbcunit framework itself is not itself thread-safe.  I've only ever looked at 'Failed == 0' before but notice now that the math doesn't work out.

```
SUMMARY

 Asserts    Passed    Failed  Suite                                      Tests

--------  --------  --------  --------------------------------------  --------
....
      18        17         0  fbc_tests.threads.threadcall_                  1
....
--------  --------  --------  --------------------------------------  --------

  632505    632504         0  Total                                       1977

unit-tests.mk:182: recipe for target 'run_tests' failed
make[1]: *** [run_tests] Error 1
make[1]: Leaving directory '/home/travis/build/freebasic/fbc/tests'
Makefile:158: recipe for target 'unit-tests' failed
make: *** [unit-tests] Error 2
The command "contrib/travis/build-and-test.sh" exited with 2.

Done. Your build exited with 1.
```

PR #258, opened 2020-10-24
Fix DX and GDI GFX drivers crashing on window creation

adeyblue:
RE: https://www.freebasic.net/forum/viewtopic.php?f=6&t=28868

Fix DirectX and GDI GFX drivers crashing when something external hooks the window (in this case an IME) and starts doing Windows message related things before we do. As these drivers show the window before they've finished initialisation, any Get/PeekMessage anybody does may cause a WM_PAINT to happen, and when it does the drivers crash /because/ they've not finished initialising.

This PR removes all the visible flags passed to fb_InitWindow and adds a ShowWindow after all the initialisation is complete.


jayrm:
Thanks @adeyblue

PR #262, opened 2020-10-25
Gas64 rev

SARG-FB:
I didn't use 'rebase --onto master start ' just 'rebase master' on gas64_rev branch. Hope no issue.

- fix sse41 test, before the asm test instruction was wrongly used.
- fix wrong optimization for xmmN registers.
- new optimizations canceling the above fix.
- some changes in asm_info() lines, typos fixes

### DIFF
--- a/src/compiler/ir-gas64.bas
+++ b/src/compiler/ir-gas64.bas
@@ -529,7 +529,7 @@ private sub check_optim(byref code as string)
                 part2=prevpart2
                 
             ''xmm register ?
-            elseif prevpart2[0]=asc("x") then
+            elseif prevpart2[0]=asc("x") and instr(prevpart1,"[")=0 then
                 asm_info("OPTIMIZATION 3")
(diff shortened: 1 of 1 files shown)


jayrm:
Thanks @SARG-FB .  Looks like that should fix my mistake with the SSE41 tests in #261 

Rebase is good.  The PR is branched from current fbc/master, so that's perfect.  It's nice to have the new changes on fbc/master as it makes picking commits for the `fbc-1.07` branch very easy.  Thank-you.

For the 4 sections beginning with `ctx.roundfloat=true` looks like there is enough duplicated code that it would be worth refactoring in to a private sub; something like `private sub hEmitRoundFloat( ... )` or similar name.  It would remove duplicate code and put the common task in one place -- if it needs to be changed in future, only have to change in one place rather than 4.  Also makes the fbc executable a little smaller.

To update this PR, you can make the changes on your PC, and when ready push to your `gas64_rev` branch again.  github will automatically update this PR and run travis-ci build tests again.

PR #263, opened 2020-10-27
gas64 : removal of duplicate code in _emitconvert by creating sub hEm…

SARG-FB:
…itRoundFloat

We must follow dkl's rules :-)


jayrm:
Yes, anytime you see more than a few lines of duplicated code in the sources is a place should consider refactoring in to a common subroutine.  The naming of the subroutine helps define what the function of the code is and where it's used.  If it needs to change, only have to change in one place instead of searching the sources for all uses.

As mentioned in other PR's, my preference is to rebase.  However, I'm going to try a simple merge on this update.

PR #269, opened 2020-11-14
Gas64 rev

SARG-FB:
I ended up to remove the use of DBGLINEEND, fixing some 'holes' hidden by reg_freeall (not used anymore).
Next step : reduce the footprint in memory of each asm procedure.


jayrm:
@SARG-FB , I see what you are going for with the PR.  I'm testing your revision and making some changes.

1) Setting the default stacksize needs to be delayed until after all of the command line arguments are parsed.   Testing fbIs64Bit() is not valid until the actual target is known, as would be the case for cross-compiling.
2) I have added an unrelated fix for debugging in symbTypeToStr() dealing with function pointers.  Previously was a failed assert when dumping the contents of an AST.  Should not change user code use of TYPEOF.
3) Moving the rtlStrAssign() special case in to rtlStrAssign() only.  I'm quite certain the assignment result is never used.  I have the feeling that there's other left over stuff from years ago with node->call.isrtl, but that's too much for me to review right now.  If the assignment result were to be used it would require many other changes to fbc syntax, optimizer & rtlib.
4) Moving the special case of ThreadCall to astNewLink() called from rtlThreadCall() using an extra parameter 'discard_results' = false, and which everywhere else currently defaults to true.

Overall, I'm willing to give this a try, but I'm not quite comfortable yet with the method for removing unused call returns.  I think we are making some assumptions about LINK & CALL nodes, unless you have really *really* looked at all the astNewLink() calls.

jayrm:
@SARG-FB 
I've given this change a lot of thought.  I have some other changes I'd like to merge in first, then I will return to this.

The overall concept is a good idea.  Optimizing out unused call returns is good for all emitters.
In general, the preferred goal should be:
- Fully construct the AST
- playback the AST to the IR/emitter without few if any replacements.

If the decision for unused return values can be made in astLoadLINK() then preferred would be to optimize it out at the time of astNewLINK().  And if that's possible, then really we would want to solve out at the time astNewCALL().  Of course, this requires many changes to the fbc source rather than solving out after the fact in astLoadLINK().

The reason we'd like to solve out unused return values earlier is that it makes the AST predictable and results known.  And for debugging when dumping the contents of an AST, we get closer to what's actually played back by IR/EMIT rather than having AST modified last moment.

I feel like the concept to remove unused returns is solid but could be solved out better.  After I make some other merges I will return to this PR.

SARG-FB:
@jayrm 
Now the 'decision' is in astNewLINK(), I understand that doing things as early as possible is better. But that is beyond my current knowledge of the compiler... So I let you think about this point.

I made some changes in gas64 : reduction of the memory footprint for every procedure and some minor others. Pushed tomorrow.
I have created and pushed a new branch  'Regroup_debug' to regroup most of elements used in internal debugging. It's still a bit raw  : compilation with -edebug is correct but I must compile without and make some tests. 

Laurent.

jayrm:
@SARG-FB, I have reworked this PR and will force push soon.  If you've made other changes and want to preserve locally, be sure to make a local branch of `gas64_rev` before you pull from your github repo.

I have re-ordered the commits, cleaned out a few unnecessary changes, and have added changes of my own.  The problems start where `reg_freeall()` is removed and having a fixed length array for the spilled registers, after which the requirement is on AST to generate code that does not exceed the limits of the emitter.  I've made changes to `ir-gas64.bas` so that the spilled vregs are now a dynamic list.  Even if AST is poorly optimized, gas64 should still allow itself to succeed in producing an executable.

Overall it's almost the same change as you had originally, plus new commits from me.  I've already run this though travis-ci and all jobs pass.  https://travis-ci.org/github/jayrm/fbc/builds/751680415

Optimizations in `astNewLink()` should now be safe (understood).  Previously we were assuming that all cases could be optimized without any study.  But because we didn't really know for sure that could potentially introduce new bugs we never had before.  I've changed 'astNewLink()' so that the AST is only optimized explicitly (by reviewing every `astNewLink()` call).  Because of that, we are likely missing a few cases that could be optimized, but we can address those in a separate update so we can make progress with this PR.

Using `./tests/fbc-tests.exe` as the test case on windows I get the following results:
- ~1% decrease in executable size for gas64 emitter
- ~1% decrease in executable size for gcc emitter
- no change on gas (x86) emitter - it does it's own optimizations

PR #270, opened 2020-11-24
Implement a ThreadSelf function

adeyblue:
Partly due to seeing the second code section here (https://www.freebasic.net/forum/viewtopic.php?f=3&t=28896#p277042) and it needing mutexes and pointers to get the thread id to the actual thread it identifies, I made a ThreadSelf function. It's a little bit embarassing for something that simple.

There's a little of bit of plumbing due to needing the id to remain unique/allocated until the created thread has exited or it has being ThreadDetached/ThreadWaited on, whichever is longer. Plus having to fake a FBTHREAD id for the main thread


jayrm:
This looks good to me.  It tested fine for me on winxp, win7, linux-aarch64.

Thank-you for writing the test suite.  It reminds me that I have a TODO to finish with console output.  I'll make a few changes to fbcunit and then merge this PR.

With the extra plumbing it feels like there is almost a solution to `threadcall` execution timing as well if a thread started flag were added.  Of course that's only part of the problem as I think would still have to mark the user's threadproc with some specifier in user code with some help from the compiler.  I don't think there is any other way to signal that the thread has started unless it originates from the user's actual procedure.  Leave that for another time.

Issue #92, opened 2018-08-03
Format(0, "0.00e+00") returns 0.00e-01


srvaldez:
should return 0.00e-00

countingpine:
Related thread: https://freebasic.net/forum/viewtopic.php?f=3&t=26921&p=250211
Adding more zeroes before the decimal point further decrements the exponent, e.g. `format(0, "0000e+0")` -> `0000e-4`.
It seems to follow the same pattern as if 0.1 were used, e.g. `format(0.1, "0000e+0")` -> `1000e-4`.

Ideally the exponent should probably always be 0.  Probably just need to make zero a special case...

countingpine:
I think I can fix this by simply preventing the line https://github.com/freebasic/fbc/blob/1.07.1/src/rtlib/str_format.c#L286 from executing when `value` is zero.
But I've found further oddities, e.g. with the format string `000e+0`:
`0.1234` -> `123e-3`
`0.0000` -> `000e-3` (or `000e+0` with the suggested fix)
`1.2340` -> `001e+0`

In the latter case it should presumably be `123e-2`, but I don't currently have quick access to VB's `Format` to compare...

srvaldez:
Hi counting_pine<https://www.freebasic.net/forum/memberlist.php?mode=viewprofile&u=220>
Im am happy that you are working on this, I made a small VB6 application to format numbers, I hope that since you don't have access to VB6 that it will be of help
https://github.com/srvaldez/FreeBASIC-projects/raw/master/Formatting.zip
________________________________

jayrm:
Fixed by #272

PR #272, opened 2020-12-14
Fix FORMAT() bugs

countingpine:
- make FORMAT() always use all the fix digits in exponential form, e.g. format(1234, "###e+#") = "123e+1"
- github #92: make FORMAT() always have an exponent of 0 for zero.


jayrm:
Looks good to me.  I think I've investigated this a few times and given up every time due the complexity when diving in to the details.  Good job.

jayrm:
Fixes github #92

PR #273, opened 2020-12-22
Process GfxDriver messages outside of  driver locks

adeyblue:
... and keep painting while sizing/moving. OpenGL does extra guff in driver lock/unlock so whether that still works well now there's extra lock points is anyone's guess. I ain't know nuthin' about no stinkin' OpenGL

### DIFF
--- a/src/gfxlib2/win32/gfx_driver_ddraw.c
+++ b/src/gfxlib2/win32/gfx_driver_ddraw.c
@@ -473,9 +473,9 @@ static DWORD WINAPI directx_thread( LPVOID param )
             }
         }
 
-		fb_hHandleMessages();
-
 		fb_hWin32Unlock();
+
+		fb_hHandleMessages();
 	}
 
 	return 1;
--- a/src/gfxlib2/win32/gfx_driver_gdi.c
+++ b/src/gfxlib2/win32/gfx_driver_gdi.c
@@ -306,9 +306,9 @@ static DWORD WINAPI gdi_thread( LPVOID param )
             }
         }
 
-		fb_hHandleMessages();
-
 		fb_hWin32Unlock();
+
+		fb_hHandleMessages();
 
 		Sleep(10);
 	}
--- a/src/gfxlib2/win32/gfx_driver_opengl.c
+++ b/src/gfxlib2/win32/gfx_driver_opengl.c
@@ -71,6 +71,7 @@ static void driver_unlock(void);
 static void driver_flip(void);
 static int *driver_fetch_modes(int depth, int *size);
 static void driver_poll_events(void);
+static void opengl_handle_keyboard(void);
 static int opengl_init(void);
 static void opengl_exit(void);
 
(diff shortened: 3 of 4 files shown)


SARG-FB:
I guess it's for fixing the bug when moving a window (under Windows) the display is frozen.
In gfx_win32.c   Line 195 lock then 198 lock again. Should not be unlock line198 ?

jayrm:
First blush this looks like the correct approach.  But, can `WM_TIMER` cause re-entrant message loop behaviour?  Probably irrelevant, see below...

My opinion is that the locking scheme for gfxlib2 is incorrect and has been for a while.  It's not clear what the original intent was and is not clear what each level of locking (function) should really be doing.  My opinion is that the locks need to be re-evaluated as to their purpose and redesigned.

Trying some examples in ./examples/graphics and I seem to get some dead-lock scenarios with this change.  I didn't investigate what driver or what examples, as I think there are deeper issues.

We have `FB_GRAPHICS_LOCK()` which maps to `fb_GraphicsLock()` which seems intent is to lock gfxlib2 overall.
We have `FB_DRIVER_LOCK()` which maps to `fb_GfxLock()` which seems intent is to lock just the update / dirty array, however, `FB_DRIVER_LOCK()` implicitly calls `FB_GRAPHICS_LOCK()`
Then typically, in user code functions (e.g. PSET)  `FB_DRIVER_LOCK()` is nested in a `FB_GRAPHICS_LOCK()`, 

Then we also have direct calls to the "update_lock" in the driver itself which don't help for understanding what needs to be locked and in what order.

So, it's a bit of a mess in my opinion.  And gets messier with OpenGL and other platforms.  Win32 has seen the most attention over the years (but is still not right in my opinion).

I'm reminded of these bug reports:
https://sourceforge.net/p/fbc/bugs/884/
https://sourceforge.net/p/fbc/bugs/885/

Ideally and minimally:
- Lock the working page for user code (LINE, CLS, etc)
- Lock visible page and dirty array when reading / writing, either from user code or the screen updater (message loop).
- Lock everything when page swapping or user explicitly locks (SCREENLOCK).
- I'm probably overlooking something here ... maybe the read-back functions like POINT, BSAVE, GET, etc.

There doesn't seem to be any facility for explicitly locking the working page or visible page though there seems to be some attempt through `DRIVER_LOCK()`.  Anyway, I'd be happy with a lock everything 

In otherwords, there seems to something wrong at the design level that needs to be fixed, rather than some bug fix.

adeyblue:
OK, that's my fault. I saw the macro DRIVER_LOCK(), saw it called a function called fb_GfxLock and stupid me thought it might just do what it says. But of course, it doesn't. It takes other locks too because nobody has a clue about the lock hierarchy involved in this thing, or what the locks are supposedly protecting, so lets just lock everything all the time so we don't have to think about it.

As for redesigning it, yes, dear god, please. Maybe gfxlib3 will be the charm. After 1.08, suspend rtlib contributions in C, finish that conversion of rtlib to FB and hey, there's the excuse to rewrite gfxlib. Just like the guy who submitted the bug this PR was for though. I'm more than happy for it to be done, but I'm not going to be the one to do it

I pushed an update that actually does just lock the driver rather than all the ancillary stuff the other one did. In the common cases (ie non GL) this is the same lock they had while processing messages, and only that lock. If it still deadlocks, bin this PR and the bug, because this architecture doesn't allow it to be fixed.

PR #277, opened 2021-01-01
Bindings updates and fixes from fbbindings

rversteegen:
This updates headers for a number of libraries, and brings fbc and fbbindings into sync: all headers that are present in fbbindings will be identical to those in fbc. Also, all fbbindings tests now pass (which is not to say that you can run "make" to successfully re-make all headers without changes).

In the process of doing this merge I found and fixed some bugs in sqlite3ext.bi and win/windowsx.bi and a small one in SDL/SDL2.

I also added a binding for SoLoud because I'm using it; it's a much nicer alternative to SDL_mixer.


jayrm:
Looks good.  Had very few issues between fbbindings and this pull request.

PR #281, opened 2021-01-10
fbc : remove dbg nodes for empty and comment lines

SARG-FB:
New try....

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -37,6 +37,7 @@ Version 1.08.0
 - fbc: double the minimum and default stacksize on 64-bit to 64Kb and 2048Kb respectively.
 - fbc: internal changes to optimize away unused call results
 - github #203: allow casts of addresses on static initializers 
+- only write debug line information for statements and don't write comments / empty lines / directives for top level source code in assembly debug ouput
 
 [added]
 - extern "rtlib": respects the parent namespace, uses default fb calling convention and C style name mangling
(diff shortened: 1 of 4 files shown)


jayrm:
@SARG-FB I made a few changes.  I guess this is what you had in mind? For top level parsing only?  Is there a reason for not writing out the comments and blank lines to ASM file?  Is it because there is now no debug node to attach to?  

Just so you know, it's still not correct: the output that is in the ASM file is wrong.  Have to take care with the parsing loops since some lex*() functions can invoke ppCheck()/ppParse() if not already in the preprocessor.   And the current line output can get buffered with the first token of the current line.

So, what this does is (sort-of):
- At the top level, a preprocessing step that skips over comments and empty lines, however preprocessor statements are still invoked
- don't write comments and empty lines to debug assembly output
- don't write debug line information for comments and empty lines

I didn't touch the many other places in the parser where we have some kind of:
```
		'' Comment?
		cComment( )

		'' emit the current line in text form
		hEmitCurrLine( )
```

IMHO, I think need this idea to remove debug nodes needs a different approach to solve.

jayrm:
@SARG-FB with the last minor change I made to correct the output of statements in assembly, I think this is OK to rebase, force push, and merge in to master.

Fair warning, I feel like this is not the last we've seen of the fbc parser oddities though ...

PR #286, opened 2021-02-09
DragonflyBSD support

CretinHo:


makefile
```diff
@@ -174,6 +174,8 @@ ifdef TARGET
     else ifneq ($(filter msdos%,$(triplet)),)
       TARGET_OS := dos
     else ifneq ($(filter freebsd%,$(triplet)),)
+      TARGET_OS := dragonfly
+    else ifneq ($(filter dragonfly%,$(triplet)),)
```
dkl, line 178: These two lines should be swapped, otherwise FreeBSD and DragonFly are mixed up, which was probably not intended here. I can do that.

dkl (approved):
Looks good, thanks for doing the work of adding it. I'll just go ahead and merge, since I don't think this should break anything else, and it's definitely going in the right direction (if not even already finished).

PR #287, opened 2021-02-10
Adding the missing bits for Solaris support

CretinHo:
I added all of the missing bits for Solaris support in the compiler using the template for FreeBSD.


dkl (approved):
As for as the DragonFly pull request - thanks for doing the work of adding it. I'll just go ahead and merge, since I don't think this should break anything else, and it's definitely going in the right direction (if not even already finished).

PR #288, opened 2021-02-12
Solaris linker fixes

CretinHo:
After these fixes are applied fbc finally could produce a working binary on OpenIndiana.

### DIFF
--- a/makefile
+++ b/makefile
@@ -453,6 +453,10 @@ ifeq ($(TARGET_OS),netbsd)
     -I/usr/pkg/include
 endif
 
+ifeq ($(TARGET_OS),solaris)
+  ALLCFLAGS += -I/usr/lib/amd64/libffi-3.2.1/include
+endif
+
 ifeq ($(TARGET_OS),darwin)
   ALLCFLAGS += -I/opt/X11/include -I/usr/include/ffi
   
--- a/src/compiler/fbc.bas
+++ b/src/compiler/fbc.bas
@@ -578,10 +578,16 @@ private function hFindLib( byval file as zstring ptr ) as string
 end function
 
 private function fbcLinkerIsGold( ) as integer
-	dim ldcmd as string
-	fbcFindBin( FBCTOOL_LD, ldcmd )
-	ldcmd += " --version"
-	return (instr( hGet1stOutputLineFromCommand( ldcmd ), "GNU gold" ) > 0)
+	'' This is needed otherwise it will wrongly pass --version into the linker on Solaris
+	'' caused the linker version to be printed everytime we compile with fbc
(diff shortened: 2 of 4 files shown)


CretinHo:
> This is awfully specific. Does Solaris have pkg-config? If so, does this work?
> 
> ```
>   ALLCFLAGS += -I$(shell pkg-config libffi --cflags)
> ```
> 
> The location of libffi is a nuisance point on many different platforms.

Almost forgot about pkg-config. Of course Solaris has pkg-config. But on my OpenIndiana, it needs to be installed before we can use. This is the output of pkg-config libffi --cflags on my OpenIndiana:

-DFFI_NO_RAW_API -I/usr/lib/libffi-3.2.1/include

Everything seems to be OK. But, wait a minute. Do you see something strange?

Our port of fbc for Solaris is 64 bit only. But OpenIndiana/Illumos is a weirded OS. They are a 64 bit only OS but the toolchain will target 32 bit by default. For example, when you type gcc -v, you will see the triplet to be i386-sun-solaris2.11! If you want it to target the correct triplet, you have to type gcc -m64 -v, then it will print x86_64-sun-solaris2.11.

The correct path to libffi should be: /usr/lib/64/libffi-3.2.1/include

I appreciate your idea of using pkg-config but it seems we can't escape from hard code the path anyway.

CretinHo:
Please have a look at it, developers. It's kind of complete now. I think no new commits is needed.

CretinHo:
Caution: this Solaris port is actually about OpenIndiana/Illumos but not Oracle Solaris. Oracle Solaris was never in mind from the beginning even though I want my port to be an umbrella to support the whole Solaris/Illumos family. This port may only works on OpenIndiana but not other Illumos distros. The Illumos distros themselves incompatible. Unlike Linux where there is still some guaranteed compatibility between distros. Just tested on Tribblix and failed to bootstrap.

jayrm:
Hi @CretinHo, I started working on testing Solaris using a VM to verify but then got side tracked with a FreeBSD build.
This appears to be follow-up for #286 and #287 and appears to be safe to add even if not yet independently verified.

I agree, some of these fix-ups or improvements will just need to monitored and improved over time.

jayrm:
Hi @CretinHo, I started working on testing Solaris to verify but got side tracked with a FreeBSD build.
This appears to be follow-up for #286 and #287 and appears to be safe to add even if not yet fully verified by another developer.
I don't believe should cause conflicts with the other pull requests on the go.

I agree, some of these fix-ups or improvements will just need to monitored and improved over time.

I notice you have some additional branches solaris-fix and dragonfly-fix that are based on this PR.  Are there additional fixes coming?

PR #289, opened 2021-02-13
gas64 - fix for future issue when using ld version 2.36

SARG-FB:
srvaldez got an error when using binutils 2.36. This fix doesn't change alot it avoids a 32bit address inside a such instruction  QWORD PTR ["+lname+"+rax*8]"

BTW there is also a warning : fbextra.x contains output sections; did you forget -T?


jayrm:
For the fbextra.x warning, if LD doesn't recognize the file as an object file format, it assumes it is a linker script.  Will try adding the '-T' option explicitly in a separate fix.

PR #294, opened 2021-03-27
Simple Generic Containers

adeyblue:
I don't have time to write any examples and I won't for a long while but as they all have a bunch of tests written for them, here's a generic List (C++ Vector/C# List), LinkedList, Queue, Stack and a non-generic BitSet that should be OK to start being battle tested.

Compiling with -exx will turn on a debug mode that'll log all container elements built during compile and at runtime spit out prints of parameters to functions and when bad things happen (like popping from an empty queue or inserting at/ indexing a negative index). It'll also call Error instead of setting Err.

Due to this and the On Error not working as intended/as I expected, the tests will crash if compiled with -exx since they obviously also test the Error conditions which corrupts/don't restore the stack when they're caught

A quick usage example

    #include "containers/List.bi"
    FBCont_DefineListOf(Long)  '' needto define each data type

    '' Then you can use it like this
    dim myList As FB_List(Long)
    '' Or probably aliasing it is better
    Type LongIterator As FB_Iterator(Long)

    '' This will work with any contains that contains Longs!
    Sub PrintContainer(ByVal iter As LongIterator)

        While iter->Advance()
            Print iter->Item()
        Wend

    End Sub

    myList.Add(789)
    myList.Insert(0, 123)
    dim pListIter As LongIterator = myList.GetIterator()
    PrintContainer(pListIter)
    Delete pListIter

(This code has been written in the message and not tested, it prob won't compile as is)

I'm not mad about the FBCont_ / FB_ define names, but that's a simple change

### DIFF
--- a/inc/containers/_concat.bi
+++ b/inc/containers/_concat.bi
@@ -1,0 +1,57 @@
+#pragma once
+
+'' Concats a comma separated list of arguments
+
+#define __MAC_CONCAT_ARG_LIST1(arg1) arg1
+#define __MAC_CONCAT_ARG_LIST2(arg1, arg2) __FB_JOIN__(arg1, arg2)
+#define __MAC_CONCAT_ARG_LIST3(arg1, arg2, arg3) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST2(arg2, arg3))
+#define __MAC_CONCAT_ARG_LIST4(arg1, arg2, arg3, arg4) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST3(arg2, arg3, arg4))
+#define __MAC_CONCAT_ARG_LIST5(arg1, arg2, arg3, arg4, arg5) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST4(arg2, arg3, arg4, arg5))
+#define __MAC_CONCAT_ARG_LIST6(arg1, arg2, arg3, arg4, arg5, arg6) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST5(arg2, arg3, arg4, arg5, arg6))
+#define __MAC_CONCAT_ARG_LIST7(arg1, arg2, arg3, arg4, arg5, arg6, arg7) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST6(arg2, arg3, arg4, arg5, arg6, arg7))
+#define __MAC_CONCAT_ARG_LIST8(arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8) __FB_JOIN__(arg1, __MAC_CONCAT_ARG_LIST7(arg2, arg3, arg4, arg5, arg6, arg7, arg8))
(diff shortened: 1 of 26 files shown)


jayrm:
I poked away on this a little.
This first little change was some minor name case fix-ups when taken over to linux - no big deal.  Which I committed to this PR.
With that done, can see that the last CI checks pass when not using '-g -exx'.

Current set-up for the CI checks is:
- one run with neither '-g' or '-exx'
- one run with both '-g' (debug info + asserts) and '-exx' (on error handling).

'-g' implies '-eassert' so asserts get turned on and fbc will just die if assert fails.  assert() doesn't throw the error.
'-exx' turns on the 'on error goto' handling.

The `ON ERROR GOTO` is generally broken.  So that sucks.  I think this one of the first times seeing it actually added to the test suite.  I looked at what it's going to take to get `on error goto` doing something reasonable ... and it's a lot.

I have another commit here that progresses the '-g -exx' a little further which moves automatic definition of FB_CONTAINER_DEBUG to helpers.bi
- otherwise appears that g_stopRecursiveLoggingFlag won't get defined
- and g_stopRecursiveLoggingFlag is needed later if FB_CONTAINER_DEBUG is defined
- g_stopRecursiveLoggingFlag appears to also be needed by FB_CONTAINER_LOG_CALLS

Which led me to believe that maybe I don't fully understand what debugging modes are expected.  It's kind of a new thing to have an implementation added to the test-suite where we both want to test the results but maybe also have a debug mode to debug the implementation itself.

It seems close, but I think need to sort out:
- fix `on error goto` or rewrite tests another way?
- add a wrapper for assert()?  otherwise with the existing test-suite framework, individual tests (expected to fail) are needed as separate test files to test a single fbc run-time assert().
- the intent of `FB_CONTAINER_DEBUG` and `FB_CONTAINER_LOG_CALLS`.  I'm not sure if these are to debug the container code itself, or the user code, or both.

adeyblue:
For the on error goto tests, is there another way? If you remove the explicit Error statements and run the ifdeffed-out tests, it's still going to crash because those Errors are in places where it's a very real error that's going to crash the program if it continues. How do you make popping an empty queue safe and/or catachable? I'd argue it needs testing because its proper to do so, but I'm very much a correct trumps easy type of person, which why I keept writing On Errors in even after finding out it didn't work.  On the other hand, I'm also very much /not/ going to be trying to make On Error Goto work correctly, so what I prefer doesn't really matter..

The rationale for turning the Err's to Error when -exx is defined is because -exx turns on FB array bounds checking, and out of bounds container access is essentially the same thing. And that's what the debug macros are for, to help answer 'how did I get to this error state?'

OK it isn't the best debug feature anybody's ever written since FB has no way of answering 'who called me' but you still get a log of the state affecting sequence of calls. -exx essentially stands for extra extra debug, and that's what it is. Verbose log of your usage of the container interface. It just seemed nicer from a user point of view that if their program crashes on a queue.Pop(), they can simply add a define and see a whole log, than having to then instrument their whole container-using code themselves or adding a bunch of tracepoints in a debugger. I got the idea from Jose's headers, blame him for it :-)

FB_CONTAINER_LOG_CALLS is entirely for being able to turn it on without exx

Either way, the weather's warming up, you've prob had your jab, so take the path of least resistance. If its causing that much consternation, rip out all the stuff that's making things difficult and move on to more important things.

PR #295, opened 2021-04-06
We need libgpm-dev to be able to build fbc on Debian

CretinHo:
I tried to build fbc on Debian and found I need this library even though it's not listed on readme.txt.


jayrm:
Perfect.  Thanks for the PR.

PR #297, opened 2021-04-09
Show warnings when FOR variable cannot exceed TO limit

countingpine:
e.g. `for i as ubyte = 0 to 255`

This seems to come up occasionally (e.g. https://freebasic.net/forum/viewtopic.php?f=3&t=29304), so I thought it might be worth having a warning for.

I'm quite rusty on compiler internals, and I fully expect that this can be approved upon, but sometimes it's perhaps easier to start by proposing something and seeing how it can be improved, than just to throw out a suggestion and add it to a mental Todo list somewhere.

It covers upper and lower limits for all the integer types, and I've started a warning test that should hit all the cases.


jayrm:
I imagine the limit check gets hairy if the step value is considered.  But it looks like a good place to start and probably warns the most common case, so I'd like to merge this in.  I'll add the warning output files and change log message.

PR #300, opened 2021-04-25
- forcing the alignment for structure in array if not a multiple at l…

SARG-FB:
…east of 4

  get a case with one winapi function (createacceleratortable) not working


jayrm:
One job failed to start, which has nothing to do with the actual change.  So, looks good to go.

SARG-FB:
Hi Jeff fast reply,  you are on the deck :-)

PR #308, opened 2021-05-06
gas64 : Fixed the signature returned in the debug information is not the signature of the currently compiled FBC source code version. 

skyfish4tb:
gas64 : Fixed the signature returned in the debug information is not the signature of the currently compiled FBC source code version. For example, if you use a compiler with a compiler signature of "freebasic 1.08.1" to compile the FBC source code with a signature of "freebasic 1.07.3", the signature returned in the debugging information of the compiled FBC source code is "freebasic 1.08.1".

### DIFF
--- a/src/compiler/ir-gas64.bas
+++ b/src/compiler/ir-gas64.bas
@@ -887,7 +887,7 @@ sub edbgemitheader_asm64( byval filename as zstring ptr )
 	dbg_addstab ("DUMMY",0)
 
 	''placeholder for information like compiler version, etc
-	dbg_addstab (__FB_SIGNATURE__,255)
+	dbg_addstab (FB_SIGN,255)
 
 	'' directory
 	if( pathIsAbsolute( filename ) = FALSE ) then


skyfish4tb:
gas64 : Fixed the signature returned in the debug information is not the signature of the currently compiled FBC source code version.

SARG-FB:
I'm not sure to understand :-)
FB_SIGN is internal ?
What is the value of FB_SIGN ?

skyfish4tb:
One is the signature of the compiler used to compile the source code, and another is the signature of the compiled source code itself.

skyfish4tb:
If the compiled source code is version 1.08, and FBC v1.06 is used to compile the source code. __FB_Signature__ is "FreeBasic 1.06"
, and FB_Sign is "FreeBasic 1.08".

SARG-FB:
Ok understood. Thanks

Btw do you study all the code or are you using gas64 ?

skyfish4tb:
Add this to the debug information to indicate which compiler version you are using?

SARG-FB:
> Add this to the debug information to indicate which compiler version you are using?
Yes the purpose is to get the version of the compiler used for compiling the debuggee, not the compiler :-)
It's an information that doesn't exist in the original Stabs (debug data). I use the code = 255

Tested on my side. It's ok.

Note : the debug data can only be used with the new version of fbdebugger, currently work in progress, as other debuggers on 64bit don't know that format (only 32bit).

skyfish4tb:
If you compile the source code of FBC v1.08 with other compilers, and then use the newly compiled FBC (v1.08) to compile other codes, the resulting signature will be incorrect.

SARG-FB:
Sorry I wanted to say Tested on my side. It's ok      now with your mod........

PR #312, opened 2021-05-18
basic-macros: add __FB_ARG_LOOPEXPAND__( ), __FB_OBJ_OFFSETOF__( ), __FB_OBJ_VTABLETYPE__( ) builtin macro.

skyfish4tb:


jayrm:
The first commit is not related to the second and third commits, so could have submitted as separate pull requests.

`__FB_ARG_LOOPEXPAND__` looks interesting - when testing seemed more like a LIST operation than a LOOP operation.  Anyway it will suffer same problems as `__FB_ARG_EXTRACT__` as reported in issue #321 .  I am working on a fix for #321 .

For the other 2 macros, I like the concept and would like to see the use case.  `__FB_OBJ_OFFSETOF__()` is very low level.  Ok maybe should be a global macro handled at the lexer level.  But it seems ` __FB_OBJ_VTABLETYPE__( )` would be better handled at the parser level as it is declaring a type in to the current namespace and I would think should only be allowed in places where TYPE declarations would be allowed.

jayrm:
I was surprised to see this pull request closed.  I think we want to do ... something with this.

PR #323, opened 2021-06-25
gas64: When the target variable is const variable, code cannot be generated for variable assignment of different length.

skyfish4tb:
gas64: When the target variable is const variable, code cannot be generated for variable assignment of different length, such as:
       dim as byte bValue = 123
       dim as const UInteger uiValue = bValue


jayrm:
Type information is kept as long as possible so that ast/ir/emit can make decisions about it if necessary.  But ultimately, yes, the extra type information not needed should be dropped once getting to the level of emitting register and memory moves, etc.

PR #324, opened 2021-06-26
When converting a wide character escape sequence to an internal format, the escaped value is incorrect.

skyfish4tb:
When converting a wide character escape sequence to an internal format, the escaped value is incorrect.


jayrm:
thank-you.  The test-suite does not well cover encoded (i.e. non ascii) fb source files.

PR #325, opened 2021-06-28
Enables __FB_EVAL__() to support wide character mode.

skyfish4tb:


jayrm:
I appreciate the effort for code reuse though it seems at the expense of complexity with the use of #define's as means for template coding.  Unless there's some other motivation for the implementation, I think it would be best just to maintain the separate string and wstring paths of execution through the lexer.

On first inspection:  the hEscape[W] overloads seem superfluous - an optional argument seems like it should work also.

I'll check this out.

skyfish4tb:
When the escape character contains a string terminator, hUnescape[W] converts the internal format to a normal string sequence, the length of the converted string may be incorrect, you may get the string length before the first terminator. Such as !"Test1\0\test2", at this time, you need to obtain the real length through optional parameters(textlen).

skyfish4tb:
When the escape character contains a string terminator, hUnescape[W] converts the internal format to a normal string sequence, the length of the converted string may be incorrect, you may get the string length before the first terminator. Such as !"Test1\0test2", gets the length of the string "test1", at this time, you need to obtain the real length through optional parameters(textlen).

jayrm:
> you need to obtain the real length through optional parameters(textlen).

The second overloaded function is not needed.  It won't break source compatibility to add an optional parameter.

```
declare function hUnescape _
	( _
		byval text as zstring ptr, _
		byref textlen as integer = 0 _
	) as zstring ptr

declare function hUnescapeW _
	( _
		byval text as wstring ptr, _
		byref textlen as integer = 0 _
	) as wstring ptr
```

jayrm:
> When the escape character contains a string terminator, hUnescape[W] converts the internal format to a normal string sequence, the length of the converted string may be incorrect, you may get the string length before the first terminator. Such as !"Test1\0test2", gets the length of the string "test1", at this time, you need to obtain the real length through optional parameters(textlen).

Thank-you for the description.   Better would be an "escape-to-fbc" type of function that can take the internally escaped string and convert it back in to a quoted string token that can be used in source code (i.e. the result passed back to the lexer).  

When the '!' prefix is added to the beginning of the string then will need to also handle backslashes '\' in addition to the NUL characters.

Here's a reference example / test-case - only using STRING.  There will be same issue with WSTRING.
We should expect a length of 3 in every case.

```
#macro dump( arg )
	scope
		print #arg
		print "  text : '" & arg & "'"
		print "  len  : " & len( arg )
		dim n as integer = len( arg  )
		dim a as string = arg
		print "  bytes:"; 
		for i as integer = 1 to n
			print " "; asc(a, i);
		next
		print
	end scope
#endmacro

dump ( "A B" )
dump ( !"A\000B" )
dump ( "A" + !"\0" + "B" )
dump ( chr(65, 0, 66 ) )
dump ( !"\\\000\\" )

dump ( __FB_EVAL__( "A B" ) )
dump ( __FB_EVAL__( !"A\000B" ) )
dump ( __FB_EVAL__( "A" + !"\0" + "B" ) )
dump ( __FB_EVAL__( !"\\\000\\" ) )
'' dump ( __FB_EVAL__( chr(65, 0, 66 ) ) ) '' error, expected constant

/'
OUTPUT:

"A B"
  text : 'A B'
  len  : 3
  bytes: 65 32 66
!"A\000B"
  text : 'A B'
  len  : 3
  bytes: 65 0 0
"A" + !"\0" + "B"
  text : 'A B'
  len  : 3
  bytes: 65 0 0
chr(65, 0, 66 )
  text : 'A B'
  len  : 3
  bytes: 65 0 66
!"\\\000\\"
  text : '\ \'
  len  : 3
  bytes: 92 0 0
"A B"
  text : 'A B'
  len  : 3
  bytes: 65 32 66
"A0B"
  text : 'A0B'
  len  : 6
  bytes: 65 27 1 48 66 0
"A0B"
  text : 'A0B'
  len  : 6
  bytes: 65 27 1 48 66 0
"\0\"
  text : '\0\'
  len  : 10
  bytes: 27 92 27 1 48 27 92 0 0 0
'/
```

Also will have to be careful of DWSTRING results in static variables since `__FB_EVAL__` can be re-entrant.
example: `print __FB_EVAL__( "A" + __FB_EVAL__( !"\000" + "B" ) )`

I understand the attempt to re-use code.  Personally, I would have written out each procedure fully.  The extra `#defines` and `#macro` is a level of indirection I would not have bothered with just to keep the code simple to read.

Anyway, I understand now the bug you are reporting here.

jayrm:
I'm going to take the approach of writing this change another way.  Posted in #326.

jayrm:
Concept of changes incorporated by PR #326

PR #327, opened 2021-07-15
When a WString contains double byte characters(such as Chinese characters), the byte size of the required character buffer is calculated incorrectly.

skyfish4tb:
When a WString contains double byte characters(such as Chinese characters), the byte size of the required character buffer is calculated incorrectly.
for example: #Print 1, "中2", 3 /' codes saved in ule16 '/


jayrm:
It would be so much easier if you would provide code to show the problem.
Somewhere else, in some code you don't show, I will guess you are trying to convert wstring to zstring but encode it as multibyte instead of ascii - which has expectation of loss during the conversion.

```
'' copy wstring to a zstring

#macro dumpBytes( x, n )
	for i as integer = 0 to n-1
		print hex( cast(ubyte ptr, x)[i], 2 ) & " ";
	next
	print
#endmacro

const s = "1中23"

dim w as wstring ptr = @s
dim b as zstring * 10
dim z as zstring ptr = @b

*z = *w

print "len(*w)   : " & len(*w)
print *w
dumpBytes( w, (len(*w)+1)*sizeof(wstring) )

print "len(*z)   : " & len(*z)
print *z
dumpBytes( z, len(*z) + 1 )

/'
OUTPUT:
len(*w)   : 4
1中23
31 00 2D 4E 32 00 33 00 00 00
len(*z)   : 4
1?23
31 3F 32 33 00
'/
```

Also, careful with the `cast( ushort ptr, src)`.  The `sizeof(wstring) <> 2` on some systems.

Issue #290, opened 2021-02-16
Porting FBC to FreeBSD on PowerPC64


lenoil98:
I'm trying to port FBC to FreeBSD on PowerPC64 using FreeBASIC-1.07.3-source-bootstrap from sourceforge and FreeBSD-x86_64.

I made what I believe are the necessary changes to the compiler source (.bas) files and created new bootstrap-dist for powerpc64.  There are a couple of issues I haven't been able to resolve. 

The first issue is when "fbc --version" is executed on the target system, the bootstrap compiler displays:

FreeBASIC Compiler - Version 1.07.3 (2021-02-16), built for freebsd-x86_64 (64bit)
Copyright (C) 2004-2019 The FreeBASIC development team.

This is not a major issue.

The major issue is when attempting to build the new fbc compiler,  using the bootstrap, everything seems to go well until the linking.  The following is displayed:

[...]

PR #336, opened 2021-08-10
Add initial support for FreeBSD on PowerPC

lenoil98:
Adds code to support FreeBSD on PowerPC. New and updated headers to provided, still a work-in-progress. 
Tested on FreeBSD 13.0 on big endian and little endian.


jayrm:
OK, I see you have just closed - so I won't bother with code review.

The changes to enable the ppc targets in the makefile fb.bas fbc.bas, etc are mostly OK.  Pay attention to *all* the details including comments and whitespace in the source code please.

The changed headers for freebsd are a bit of a mess.  duplicate declarations, incomplete translations, etc.  #ifdef __FREEBSD__ blocks  with no explanation for disabled code.

A few weeks ago, I completed full tests on freebsd 13.0 for i386 and amd64 passing the entire test-suite.  Even though crt headers are incomplete, I added just enough needed to compile the compiler, bootstrap compilers, and test-suite + framework.  It's basically ready for a release.

inc/crt/lnetdb.bi
```diff
@@ -0,0 +1,160 @@
...
+#include once "crt/netinet/in.bi"
+#include once "crt/stdint.bi"
+''
+''
+''
+''
+''
+''         be included in other distributions without authorization.
+'' NOTICE: This file is part of the FreeBASIC Compiler package and can't
+'' bits\netdb -- header translated with help of SWIG FB wrapper
```
rversteegen, line 87: What's going on here? Where have these headers come here? Copy-pasted from other headers, with or without hand edits? Are any of these new headers actually machine-translated, or these NOTICEs all copy-pasted?

src/compiler/fbc.bas
```diff
@@ -3114,10 +3114,18 @@ private function hCompileStage2Module( byval module as FBCIOFILE ptr ) as intege
...
 
 		if( fbGetOption( FB_COMPOPT_TARGET ) <> FB_COMPTARGET_JS ) then
-			if( fbc.cputype_is_native ) then
-				ln += "-march=native "
+			if( fbGetCpuFamily( ) = FB_CPUFAMILY_PPC or FB_CPUFAMILY_PPC64 ) then
+				if( fbc.cputype_is_native ) then
+					ln += "-mcpu=native "
+				else
+					ln += "-mcpu=" + *fbGetGccArch( ) + " "
+				end if
```
rversteegen, line 3122: Firstly, `FB_CPUFAMILY_PPC or FB_CPUFAMILY_PPC64` is a mistake.

I was surprised by this switch to -mcpu, so I looked into it. I am shocked and appalled to find that the meaning and availability of the `-march`/`-mtune`/`-mcpu` arguments to gcc are target-specific. The PPC target doesn't even recognise `-march`, and they are especially complicated on ARM. Clang [tries to copy](https://groups.google.com/g/llvm-dev/c/oNN5Ob2Dc6I/m/j3i1iiadCwAJ) GCC's insane argument scheme, but other LLVM utilities do not.

It would be very helpful to have a note about that in a comment.

rversteegen: Also, `-mcpu=... -mtune=...` would be more closely equivalent to the `-march=...` used for other targets. [E.g. this](https://github.com/mfem/mfem/issues/216).

lenoil98: @rversteegen Good to hear from you.  These are your headers, that you pointed me to.  I apologize for not discussing this with you.  Some cut and pastes were made in response to real world testing that I mentioned in #290.

inc/crt/lnetdb.bi
lenoil98: @rversteegen Good to hear from you. These are your headers, that you pointed me to. I apologize for not discussing this with you. Some cut and pastes were made in response to real world testing that I mentioned in #290.

rversteegen: Hi; sorry for disappearing so long.

Wow, you're right, these totally butchered changes to the include files are my fault! After a lot of head scratching, I see that  `inc/crt/fnetdb.bi` and `inc/crt/lnetdb.bi` are not real headers. They're just garbage versions of `netdb.bi` that I was I using to diff, which I checked into to git accidentally in commit https://github.com/rversteegen/fbc/commit/8082ee67320527ddeba8b443de89874255f8ba6f, because I didn't give my changes even a cursory glance. These header changes were definitely not meant to be merged; they need a lot of work. The other changes to the three `netdb.bi` files may be alright, but I wouldn't count on it. There are also many other bad include file changes.

I suggest that you remove all include file changes from this PR, and we start over on them.

Now that Jeff has started some work on FreeBSD include files I think it's preferable to continue the way he's going, which is to have separate versions of each header for each OS, rather than a single file with sections gated with `#ifdef`s (as the fbbindings 'crt' branch does, using fbfrog). This avoids needing to download headers for *all* OSes and run a huge fbfrog job to update anything.

src/compiler/fbc.bas
rversteegen: Ah, so you did. Thanks for adding the comment.

Also, the mistake I was referring to was
``if( fbGetCpuFamily( ) = FB_CPUFAMILY_PPC or FB_CPUFAMILY_PPC64 ) then``
(which github unfortunately has cut out of the preview on the Conversation tab) should be
``if( (fbGetCpuFamily( ) = FB_CPUFAMILY_PPC) orelse (fbGetCpuFamily( ) = FB_CPUFAMILY_PPC64) ) then``

inc/crt/lnetdb.bi
lenoil98: Agreed. I’ll remove the headers from this PR.

src/compiler/fbc.bas
lenoil98: I’ll make the changes and resubmit the PR.

lenoil98:
Due to errors when syncing in the patches from my local host, I’m going to close this PR.  I will reopen this PR or create a new one after correcting errors and verification.

inc/crt/lnetdb.bi
lenoil98: All headers and changes to headers have been removed.

src/compiler/fbc.bas
lenoil98: Comment added to code explaining reason for change

lenoil98:
Reopened PR after verifying code and testing.

jayrm:
@lenoil98, I'm OK with this update.  There's a couple of nit-pick's I could make but I'll fix them up after I squash and merge (I don't think it will be helpful to preserve the complicated merge history leading up to this PR).

PR #337, opened 2021-08-10
Update fb.bi

lenoil98:
Adds code to support FreeBSD on PowerPC. New and updated headers to provided, still a work-in-progress. 
Tested on FreeBSD 13.0 on big endian and little endian.


jayrm:
Appears that this was intended to be a patch for #336 - not fbc/master
This patch currently missing from #336 

Anyway, between #336 and this PR, `src/compiler/fb.bi` is still missing something like:
~~~
const FB_DEFAULT_CPUTYPE_PPC 	= FB_CPUTYPE_PPC
const FB_DEFAULT_CPUTYPE_PPC64  = FB_CPUTYPE_PPC64
~~~

jayrm:
See comments in #336

PR #340, opened 2021-08-22
Add 'powerpc64le' arch and fixes for bootstrapping freebsd-powerpc*

jayrm:
Finally got around to testing the build packages this month and noticed a break in building the bootstrap package.  I haven't done any testing on the powerpc builds themselves, but this update should help getting them ported, so I'd like to merge this in anyway.

Issue with the bootstrap package was due to `-arch` expecting `powerpc|powerpc64` but the `-target` in the bootstrap using `freebsd-ppc|freebsd-ppc64`

In this update:
- add `powerpc64le` arch for powerpc 64-bit little endian
- normalize on `-arch powerpc|powerpc64|powerpc64le` for arch cpu-type (which is also used to determine directory naming in some cases)
- allow short-cuts of `-target <os>-ppc|<os>-ppc64|<os>-ppc64le` to specify `<os>-<arch>` (which normalizes as above)
- In the makefile, should also be able to use the `ppc*` short-cuts to normalize on `powerpc*` - note, not well tested.
- `__FB_BIGENDIAN__` is now intrinsically defined if arch is `powerpc|powerpc64`
- objinfo now sets and checks the endianness of ELF files - previously assumed was little endian only.  This fix may help solve some linking errors on powerpc* big endian mode - not yet tested though.

Unrelated to powerpc but in this update anyway:
- `-gen gas64` help option was missing from `fbc -help` output (but was in the fbc.1 man page)
- as part of an on-going and gentle effort to sanitize the whitespace in the compiler source, files changed in this PR got the `contrib/tools/check-whitespace.bas` treatment.


lenoil98:
@jayrm I'd noticed the issue with the CPU names in the makefile, but was waiting to see what else I found before submitting.  Hopefully the BIGENDIAN intrinsic define will help with some of the unit test issues.

PR #368, opened 2022-06-04
Fix mangling involving toplevel CONST and abbreviations

dkl:
This attempts to fix a mangling problem with `BYVAL x AS CONST TYPE` parameters. This toplevel CONST should be ignored in the mangling, just like it's ignored in a C++ function signature. fbc already correctly skipped the `K`, but it should also not add an abbreviation for this.

Checking gcc (on x86_64 linux):
```
void f(      int, int &, int &) { } // _Z1fiRiS_
void g(const int, int &, int &) { } // _Z1giRiS_
```
No `K` is added, and the abbreviation used for the 3rd parameter is `S_`, which is the first abbreviation, meaning that none was added for the first parameter. Also, note that C++ considers both signatures equal.

### DIFF
--- a/src/compiler/symb-mangling.bas
+++ b/src/compiler/symb-mangling.bas
@@ -475,24 +475,23 @@ sub symbMangleType _
 
 	'' const?
 	if( typeIsConst( dtype ) ) then
-
-		'' The type has some CONST bits. For C++ mangling we remove the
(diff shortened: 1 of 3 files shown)


jayrm:
Looks good!  Thanks.  I can update changelog.txt with some other near future commits.

PR #371, opened 2022-06-12
Fix bad code generation if asm ast nodes have newlines in them

adeyblue:
asm_info only comments the first line output. If the text to output has newlines in it, further lines are emitted uncommented. This is especially bad in _emitasmline - the asm_info comments the frst line of assembly, while subseqent ones are emitted into the code stream, and then all the lines are emitted into the code stream again by the asm_code.

This switches asm code to use gas' multiline comments rather than the single so subseqent lines remain commented.

### DIFF
--- a/src/compiler/ir-gas64.bas
+++ b/src/compiler/ir-gas64.bas
@@ -128,9 +128,22 @@ rbp-y -->  local vars
 	end if
 #endmacro
 declare sub cfi_windows_asm_code(byval statement as string)
+declare sub hwriteasm64( byref ln as string,byval opt as integer=KDOALL)
 
 #if __FB_DEBUG__ <> 0
-	#define asm_info(s) hWriteasm64("# "+s)
+#include once "crt/string.bi"
+
+	private sub asm_info( byval s as string )
+		dim as zstring ptr sdata = strptr( s )
+		dim as const zstring ptr newlines = @!"\r\n"
+		dim as zstring ptr iter =  strpbrk( sdata, newlines )
+		while iter
+			*iter = asc("#")
+			iter = strpbrk( iter + 1, newlines )
+		wend
+		hWriteasm64("# "+s)
+	end sub
+
 #else
 	#define asm_info(s) rem
 	#define typeDumpToStr(a,b) " "+str(a)
@@ -322,7 +335,6 @@ declare sub _emitdbg(byval op as integer,byval proc as FBSYMBOL ptr,byval lnum a
 declare sub check_optim(byref code as string)
 declare sub _emitconvert( byval v1 as IRVREG ptr, byval v2 as IRVREG ptr )
 declare function hgetdatatype_asm64 (byval sym as FBSYMBOL ptr,byval arraydimensions as integer = 0) as string
-declare sub hwriteasm64( byref ln as string,byval opt as integer=KDOALL)
 declare sub _emitvariniend( byval sym as FBSYMBOL ptr )
 declare sub _emitvarinipad( byval bytes as longint )
 declare sub _emitvariniwstr(byval varlength as longint,byval literal as wstring ptr,byval litlength as longint)


SARG-FB:
There is also a big problem : almost all, if not all, optimizations of asm code are disabled. And it's less readable.
However easy fixed.

Could you show an example with the issue with newlines.

Below an example of a missing optimization :

#O2mov r11, rax
           # call fb_ASC / mang=fb_ASC
           # symbdump=proc shared accessed declared rtl hasalias FB_ASC stdcall [ulong]
           # vr=reg 5 [ulong] symbdump=<NULL>
           # variadic=false level=1
           # stkcopy=176
           # arg vr4 [string ptr]
           # arg=reg 4 [string ptr] symbdump=<NULL> vreg=4
           # virtual register =4 real register=r11
           # KPARAM=0
   #O2mov rcx, r11
   mov rcx, rax #Optim 2
  ===================================== 
     mov r11, rax
   /* call fb_ASC / mang=fb_ASC*/
   /* symbdump=proc shared accessed declared rtl hasalias FB_ASC stdcall [ulong]*/
   /* vr=reg 5 [ulong] symbdump=<NULL>*/
   /* variadic=false level=1*/
   /* stkcopy=176*/
   /* arg vr4 [string ptr]*/
   /* arg=reg 4 [string ptr] symbdump=<NULL> vreg=4*/
   /* virtual register =4 real register=r11*/
   /* KPARAM=0*/
   mov rcx, r11

adeyblue:
This causes the original code gen problem.
https://github.com/adeyblue/fbc/blob/e952d7ea7b1bc8741122c837ee82fd8f573bd851/src/compiler/unwind-ast.bas#L467
The asm_info comments out the 'mov rsp...' since it's on the first line but the 'ret' after the newline gets put onto the next line of the file uncommened. Then both lines are output normally by the asm_code.
So instead of generating

```
#mov rsp, rbx
#ret
mov rsp, rbx
ret
```

It generates 

```
#mov rsp, rbx
ret
mov rsp, rbx
ret
```

Which caused my code to crash, which is how I noticed it
Any asm node in the AST with a newline in its text should do it though.

Substituting the comment character seemed the easiest, least invasive way to fix it. Not sure how it's caused a compiler crash in travis. Looks like there's a place that reads or checks for comments and it's crashing because there's no # any more:

> Aborting due to runtime error 7 (null pointer access) at line 809 of src/compiler/ir-gas64.bas::HWRITEASM64()

Never happened locally, I've built the compiler several times with this change in debug mode on Windows and Linux without problems.

Anyway, if you want to fix this however you want to fix it instead, I'll close and delete this.

SARG-FB:
Thanks for your answer.
I can't really follow your code. Just curious, is the asm code generated outside irgas64 ?

I guess it's not possible to add directly # inside the 'multiline'

Not sure it fixes the problem but a proposition :

A function to add # if there are NEWLINE

private sub addcomment(byref lne as string)
	static as integer lgtnl=len(NEWLINE)
	dim as integer posnl=instr(lne,NEWLINE),bg=1
	dim as string nline
	if posnl=0 then exit sub
	while posnl<>0
		nline+=mid(lne,bg,posnl-bg+lgtnl)+"#"
		bg=posnl+lgtnl
		posnl=instr(bg,lne,NEWLINE)
	Wend
	nline+=mid(lne,bg)
	lne=nline
End Sub

private sub hwriteasm64( byref ln2 as string,byval opt as integer=KDOALL)
dim as string ln,lname
#if __FB_DEBUG__ <> 0
	if ln2[0]=asc("#") then
		addcomment(ln2)   ' <---------------------------------
		ln="           "+ln2
	else
#endif
............
====================
for testing

#ifdef __FB_LINUX__
	#define NEWLINE chr(10)
#else
	#define NEWLINE chr(13)+chr(10)
#endif
private sub addcomment(byref lne as string)
	static as integer lgtnl=len(NEWLINE)
	dim as integer posnl=instr(lne,NEWLINE),bg=1
	dim as string nline
	if posnl=0 then exit sub
	while posnl<>0
		nline+=mid(lne,bg,posnl-bg+lgtnl)+"#"
		bg=posnl+lgtnl
		posnl=instr(bg,lne,NEWLINE)
	Wend
	nline+=mid(lne,bg)
	lne=nline
End Sub

dim as string strg="#titi"+NEWLINE+"toto"+NEWLINE+"tata"
addcomment(strg)
print strg
print "================"
strg="#titi"+NEWLINE+"toto"
addcomment(strg)
print strg
print "================"
strg="#titi"
addcomment(strg)
print strg
sleep

SARG-FB:
"take two" 
Sorry not clear for me, what do you mean ?

Another thing I don't understand is how do you get that multiline without having the possibility to put # on each line ?

Maybe instead replacing LF/CR by # just replace by an empty string (or an explicit text) if there is at least one LF/CR, faster and simplier.

Please could you show an example of the final result, I want to check if no problem when debugging.

jayrm:
@SARG, "take 2" meaning second attempt.  adeyblue  is using fbc's `astNewASM()` to inject / build multiline ASM directly from fbc.  ASM nodes with newlines is not something that would occur in user source.  The bug only shows up in gas64 when debug info is enabled and creating asm nodes with newlines directly from compiler side.

@adeyblue, related: `astNewASM()` is expecting input to already be tokenized.  I don't know, maybe will have to mind the new lines in a cross compile situation (even if rare) because a host encoded `!"\n"` in fbc could be different from the target fbc.  Not sure if mixed line endings would cause trouble for any tools.  I think should use `NEWLINE` constant as defined in `src/compiler/fb.bi`

@SARG, refactoring `asm_info()` in to a procedure call like adeyblue started here is probably useful for organizing intent.  It was also one of my first steps when I looked at this a bit last week.  And then I started to further separate the `__FB_DEBUG__` conditionals out of `hWriteasm64()`.

I have a makefile change of my own I'd like to add to control when gas64 debug info is enabled / generated (sometimes I want to debug fbc but produce same asm files as release versions).

So, I'm going to go ahead and merge this in to allow interests to progress and we can work together to sort out any troubles.

PR #376, opened 2022-06-20
Add support for FB exceptions / object unwinding

adeyblue:
I think the previous PRs are all the standalone improvements to pick out from this work, so that leaves this blob.
This is mainly here so I can see how many changes there are with the winning formula having being through about 5 different implementations, and to run the travis tests since it found a crashing problem last time.

Things that need resolving before this is merged:
- In C++ terms, the On Error Goto <label> statement now acts as the start of a try{} block, with the label acting as the catch() for any generated errors. Like try{} blocks, On Error Goto can now be nested in functions for far more localised error trapping rather than each On Error replacing a single global 'catch'. In the new way, what does On Error Goto 0 do? As is, it acts as the closing brace of the current try{} but all previous try{} regions in the function are unaffected, should it instead act as the closing brace to all previous.try{} regions?
The non-unwind behaviour of On Error Goto 0 is to turn off handling of errors until a new On Error is  encountered, and curtailing all previous regions would be more analogous to that. hCurtailCatchRegions in unwind-ast.bas can already work both ways, so this wouldn't be a major change

- The largest source or Error statements that anybody will encounter in the medium term will come from rtlib things like file opens etc, In order for the new mechanisms to work, fb_ThrowError in libfbexcept needs to be used to kick off the stack unwinding. Can rtlib be built in such a way that fb_ErrorThrowAt can do either the current behaviour or the new depending on how the exe/so is compiled? What if a static library is built with one behaviour and the exe the other? Code built for one system won't see Errors raised by the other

- Where do you want libfbexcept to live? It's such a hard dependency for code built with -z objunwind it's basically welded to the data the compiler produces, but apart from any bugfixes it's not likely to change much. It's a namespaced static library written in FB so it can easily have its contents moved 'in-house' and be done away with as a separate entity, having its .o files folded into the main rtlib.

- There's probably mountains of other things too


jayrm:
Pretty sure the [failed Travis-CI builds](https://app.travis-ci.com/github/freebasic/fbc/builds/252147689) on gas 32-bit in the initial PR are due to a bug that was fixed in fbc 1.06.  See commit [2e106bc05ac7310a881894527953af09de11c69a].  Easiest fix would be to bootstrap fbc from a new version in [contrib/travis/bootstrap-settings.sh](https://github.com/freebasic/fbc/blob/master/contrib/travis/bootstrap-settings.sh) . I tried fbc-1.06, which failed, and fbc-1.07 which passed.  So ... easiest fix here would be to up the bootstrap version to fbc-1.07.0.  Honestly, I didn't spend any time trying to understand why fbc-1.06 fails the boot strap because 1) fbc-1.07 works and is newer, but not too new, and 2) I won't bother to release a patched 1.06 to fix whatever the problem might be.  _{I was tempted to take the fbc-1.06 bootstrap off of sourceforge and github, but I currently use it to bootstrap FreeBSD builds.  I'll push the minor change here momentarily and we should see all green checkmarks on this PR as-is.}_

SARG-FB:
@adeyblue
I removed the asm_info call causing issue with mulilines in sub _emitasmline and also sub asm_info you added.
So please don't add that code again.

jayrm:
> * Where do you want libfbexcept to live?

I think in src/except would be a suitable home.  For 2 reasons:
- in fbc's development tree so that changes to fbc and libfbexcept can be committed at the same time so there's no disconnection of fbc code generation and libfbexcept requirements
- separate enough that can manage the build of libfbexcept's object files as an individual task (and later generation of bootstrap sources, etc) but also allowing fbc's makefile to meld libfbexcept in to fbrtlib

PR #377, opened 2022-06-23
Fix GPM support for GETMOUSE under TERM=linux

dkl:
The GPM support didn't work at all, because it looked for `libgpm.so.1`, while a current Debian/Ubuntu has `libgpm.so.2`. It looks like that was changed around 2008, so apparently it has been broken for quite a while.

Due to an ABI compatibility issue (see comments in the code) with different versions of `libgpm.so.1`, it seems best to drop support for that completely. For `libgpm.so.2` I didn't find any ABI issues.


dkl:
That makes sense; I'm guessing it would be enough to move io_mouse.c to the unix sub-directory and remove all the dummy implementations. And maybe check HOST_LINUX to auto-disable the GPM parts.

PR #379, opened 2022-06-25
Unix rtlib: Fix tty state cleanup in case of parallel shell()

dkl:
This fixes a bug with `fb_hInitConsole()/fb_hExitConsole()`. `fb_hInitConsole()` can be called repeatedly, without `fb_hExitConsole()` in between. An example for this is a multi-threaded FB program with parallel SHELL calls. But this affects other functions/keywords too, such as EXEC, DYLIBLOAD, etc. In a case like that, `fb_hInitConsole()` should only capture the original tty state (for restoring it on exit) once. It should not re-capture again and again, because then it would only capture the state it set itself before, causing the wrong state to be restored on exit.

See the commit messages for some more details. I'm relatively (but not 100%) confident that this shouldn't break anything. I've been using this locally for a couple of days, and I've added a test program which runs several tests and fails before the patch but not anymore afterwards. Thus I think this should be safe to merge.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -67,6 +67,7 @@ Version 1.10.0
 - rtlib: multikey was missing VK_NUMPAD5 and SC_CLEAR scancodes in the key table, causing NUMPAD-5 to be handled as escape in the windows D2D gfx driver
 - sf.net: #961: rtlib/gfxlib: COLOR() function returns LONG but should be ULONG
 - The DISABLE_GPM build option for the Unix rtlib now only disables GETMOUSE() for TERM=linux, which is the case that uses GPM, however GETMOUSE() for TERM=xterm now keeps working under DISABLE_GPM, since that case does not use GPM.
+- Unix rtlib: Multi-threaded use of shell()/exec()/dylibload()/dylibfree() can no longer cause input/output terminal configuration cleanup at program exit to fail
 
 
 Version 1.09.0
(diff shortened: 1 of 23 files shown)


jayrm:
Compiling the tests for older g++ (before g++ v11, I think) needs "-std=c++17" or "-std=gnu++17".  I didn't test oldest g++ that will work, but maybe g++ v5 or later supports the c++17 option.  Which is what is needed for current travis-ci.

Adding tests to `contrib/travis/build-and-test.sh`:  I think `$(uname)` could be checked since `build-and-test.sh` can only be run in a unix-y type shell and would allow `contrib/travis/build-and-test.sh` to succeed even on non-unix.

Personally, I never run `build-and-test.sh` directly (on dos/win), so an extra makefile target may be a simple enough probe to check that the fbc in use is for unix:
In `makefile` add as a dependency of `tester` (and probably other binaries too):
```
check-fb-unix: check-fb-unix.bas
	$(FBC) $(ALLFBCFLAGS) $<
```
check-fb-unix.bas source:
```
#ifndef __FB_UNIX__
#error
#endif
```
The makefile should fail with an error if fbc is not targeting unix.

Aside from above, looks good and is acceptable to me merging in this PR and experimenting with testing as a separate effort.

dkl:
Ah true, I'll add the `-std=c++17` to be safe.

Yea, I've thought about how to integrate this kind of test. Making it explicitly Unix-only sounds good.

Adding it to the CI build would also be nice, but chances are it can't work there. If I remember correctly, stdin had to be redirected to /dev/null on TravisCI, which would prevent this test from working too, since the relevant code in the rtlib isn't even triggered if `isatty()` returns false.

jayrm:
I'm fine with merging this in for now and dealing with automation as a future task.  The only way to really get this tested is to merge in and get feedback from users.

PR #381, opened 2022-07-10
Gas64 rev

SARG-FB:


jayrm:
In future, showing the changes to `src/compiler/ast-node-uop.bas` for your development aren't really necessary since all the changes are ultimately to `src/compiler/ir-gas64.bas` only.  But OK, will merge shortly.  Thanks.

SARG-FB:
Sorry, you are right, not very clever.... I'll do differently next time. Thanks for the merge.
BTW you are back with a lot of work ;-)

I'm preparing a code about the fixed len string issues.

Issue #382, opened 2022-07-11
unit-tests stuck on ascii


OPNA2608:
On Linux, I'm running roughly the following inside of a WIP package script:

```sh
# fetch bootstrapped 1.09.0 tarball, unpack, enter, apply patches

# build
make bootstrap -j${cores}
make FBC="${PWD}/bin/fbc -i ${PWD}/inc" -j${buildCores}

# test (ignoring other test suites for this example)
make unit-tests UNITTEST_RUN_ARGS=--verbose -j${testCores}

# install
make install prefix=/bla/blub
```

This works fine until `make unit-tests`. It builds the tests and starts running them, but it gets stuck on:
```
  fbc_tests.udt_wstring_.midstmt
    ascii
```

After letting it sit on this for more than an hour, it never advanced further. [Here is the full build log](https://github.com/freebasic/fbc/files/9084431/fbc-1.09.0.log).

I'm building against NixOS unstable. I tried to reproduce this on Ubuntu 22.04 interactively but failed. [Installing Nix on Ubuntu 22.04](https://nixos.org/download.html#nix-install-linux) and trying to build the package with `unit-tests` enabled also gets stuck.
```sh
git clone --depth 1 --branch init/fbc https://github.com/OPNA2608/nixpkgs
cd nixpkgs
vim pkgs/development/compilers/fbc/default.nix
# comment/uncomment the checkTarget lines, uncomment preCheck, comment out postCheck
nix-build . -A fbc --cores 4
```

Please let me know if there's anything else I can do for you to debug this.

jayrm:
@OPNA2608 , perhaps have a look at [contrib/travis/build-and-test.sh](https://github.com/freebasic/fbc/blob/master/contrib/travis/build-and-test.sh) and try:
```
make unit-tests UNITTEST_RUN_ARGS=--verbose -j${testCores} </dev/null
```

My amateur explanation is that fbc executables mess with the console and redirecting some commands to `/dev/null ` prevents these kinds of hangs.

To be honest, I didn't set-up a test to try and replicate; besides, @dkl probably has better knowledge and experience with this problem.

OPNA2608:
Doesn't seem to help, whether I do it via `make unit-tests` or directly with `fbc-tests`:
```
bt1cn@NixTop:/dev/shm/FreeBASIC-1.09.0-source-bootstrap/tests
↪ time ./fbc-tests </dev/null
^C
________________________________________________________
Executed in  105.53 mins    fish           external
[...]

PR #384, opened 2022-07-18
tests: Fix stack corruption in mid() unit tests

dkl:
1. Pass correct object to CLEAR (its first parameter is object BYREF, not address BYVAL)
2. Fix out-of-bounds store when manually setting null terminator

Fixes: 135f364274


OPNA2608:
Can confirm this fixes #382 for me.

PR #392, opened 2022-08-13
fbc: add support for SOURCE_DATE_EPOCH environment variable

jayrm:
- add support for `SOURCE_DATE_EPOCH` : https://reproducible-builds.org/docs/source-date-epoch/
- SOURCE_DATE_EPOCH is an environment variable that can control dates and times embedded in compiled sources
- If set, use SOURCE_DATE_EPOCH for intrinsic `__DATE__`, `__DATE_ISO__`, `__TIME__` macros
- This in turn controls the `__FB_BUILD_DATE__` and `__FB_BUILD_DATE_ISO__` macro built-in to the compiler at the compiler's build time
- Need to compile fbc twice to get this working: first time to add the new behaviour, second time to use it
- see src/compiler/fb.bi, src/compiler/symb-define.bas, fbc/compiler/fbc.bas


OPNA2608:
Tested with patch applied to 1.09.0, `SOURCE_DATE_EPOCH` values of

- `315532800`
- `0`
- none (`unset SOURCE_DATE_EPOCH`)

, compiling bootstrap -> patched -> patched and checking `fbc --version` output

- `1980-01-01`
- `1970-01-01`
- `2022-08-14`

This should be enough for my purposes.

I also tested some values that should be rejected by the build if you were to strictly follow the [variable specifications](https://reproducible-builds.org/specs/source-date-epoch/), though I'm impartial to that because even some of the code examples on the website are lax with this stuff.

>  If the value is malformed, the build process SHOULD exit with a non-zero error code. 

- `InvalidEpochNumber` (not a number) -> `1970-01-01`
- `315532800.555` (not an integer) -> `1980-01-01`

The elaborate C example code also rejects negative values, otherwise valid values with trailing garbage and values that exceed some datetime API relevant data type.

jayrm:
Latest modifications check for malformed values and should finish this PR.  Will merge in soon. The fbc build date (what is shown in fbc -version) is ultimately determined by the `__DATE_ISO__` built-in macro at fbc's compile time. 

Using the following test source file will indirectly test fbc's eventual handling of the build date in the version information: 
`print __DATE_ISO__ & " " & __TIME__`

Examples of valid values:
- `SOURCE_DATE_EPOCH=0 -> 1970-01-01 00:00:00`
- `SOURCE_DATE_EPOCH=315532800 -> 1980-01-01 00:00:00`
- `SOURCE_DATE_EPOCH= -> 2022-09-18 11:54:30`
- `SOURCE_DATE_EPOCH=253402300799 -> 9999-12-31 23:59:59`

Examples of malformed values (returns an error message and compilation will fail): 
- `SOURCE_DATE_EPOCH=253402300800`
- `SOURCE_DATE_EPOCH=InvalidEpochNumber`
- `SOURCE_DATE_EPOCH=315532800.555`
- `SOURCE_DATE_EPOCH=-315532800`

Error message on malformed value:
- `test-epoch.bas(1) error 328: Malformed SOURCE_DATE_EPOCH environment variable in 'print __DATE_ISO__ & " " & __TIME__'`

@OPNA2608, thank-you for the feed back.  I read again the variable specifications and agree; build process SHOULD exit on malformed values because without the checks the alternative is an undesirable silent success.  However, even though the build process itself (i.e. makefile) doesn't check validity of `SOURCE_DATE_EPOCH`, I am satisfied that compilation of fbc will fail on a malformed value which in turn fails the build process.  Overall, my opinion is that we should expect the builder to correctly set `SOURCE_DATE_EPOCH` upstream rather than add any unneeded checks to fbc's build process (i.e. makefile).

I should probably add a little documentation to our project somewhere on this.  Once I find where that will go, I'll merge this in.

PR #397, opened 2022-09-25
fbc: allow inner (scoped) types and unions

jayrm:
This change allows TYPEs and UNIONs to be declared inside of other TYPEs and UNIONs.  The main purpose is to scope a structure inside the namespace of another structure.  Secondarily, this should also generate a compatible c++ name mangling for nested types and parameters of member procedures.

For example:
~~~
type T
	n as integer
	type A
		i as integer
	end type
end type
dim x as T.A
~~~

This style of declaration does only specifies scope - it does not declare fields/members.  To declare field/member of the inner type within the outer type, a member must be named.  For example:
~~~
type T
	n as integer
	type A
		i as integer
	end type
	member as A
end type

dim x as T
x.member.i = 1
~~~

Also allowed is the declaration of static / non-static member procedures, extended types, and so on with the following rules:
- the definition of procedures must appear outside of the outermost type
- inner types may not extend an outer type but may extend an unrelated type, for example:
~~~
type T extends object
	type A extends object
	end type
end type
~~~

Private/Protected inner types are not well supported and will require further work to complete.  Since private/protected access is a compile time check only, generated code and overall name lookup should be valid.  However, more testing will be required in future.

### DIFF
--- a/src/compiler/parser-decl-struct.bas
+++ b/src/compiler/parser-decl-struct.bas
@@ -86,8 +86,8 @@ private sub hTypeProtoDecl _
 
 	select case( tk )
 	case FB_TK_SUB, FB_TK_FUNCTION, _
-	     FB_TK_CONSTRUCTOR, FB_TK_DESTRUCTOR, _
-	     FB_TK_OPERATOR, FB_TK_PROPERTY
+		FB_TK_CONSTRUCTOR, FB_TK_DESTRUCTOR, _
+		FB_TK_OPERATOR, FB_TK_PROPERTY
 		lexSkipToken( LEXCHECK_POST_SUFFIX )
 
 		cProcHeader( attrib, pattrib, is_nested, _
@@ -233,7 +233,7 @@ private sub hFieldInit _
 	'' constructor, if no constructor was specified.
 	symbSetUDTHasInitedField( parent )
 
-    '' ANY?
+	'' ANY?
 	if( lexGetToken( ) = FB_TK_ANY ) then
 		'' don't allow var-len strings
 		if( symbGetType( sym ) = FB_DATATYPE_STRING ) then
@@ -500,7 +500,7 @@ private sub hTypeMultElementDecl _
 		byval attrib as FB_SYMBATTRIB _
 	)
 
-    static as FBARRAYDIM dTB(0 to FB_MAXARRAYDIMS-1)
+	static as FBARRAYDIM dTB(0 to FB_MAXARRAYDIMS-1)
 	dim as zstring ptr id = any
 	dim as FBSYMBOL ptr subtype = any
 	dim as integer dtype = any, bits = any, dims = any, fieldattrib = any
@@ -540,7 +540,7 @@ private sub hTypeElementDecl _
 		byval attrib as FB_SYMBATTRIB _
 	)
 
-    static as FBARRAYDIM dTB(0 to FB_MAXARRAYDIMS-1)
+	static as FBARRAYDIM dTB(0 to FB_MAXARRAYDIMS-1)
 	dim as zstring ptr id = any
 	dim as FBSYMBOL ptr subtype = any
 	dim as integer dtype = any, bits = any, dims = any
@@ -750,7 +750,7 @@ private sub hTypeBody( byval s as FBSYMBOL ptr )
 
 	do
 		select case as const lexGetToken( )
-        '' visibility?
+		'' visibility?
 		case FB_TK_PRIVATE, FB_TK_PUBLIC, FB_TK_PROTECTED
 			if( symbGetUDTIsUnion( s ) ) then
 				errReport( FB_ERRMSG_SYNTAXERROR )
@@ -776,7 +776,7 @@ private sub hTypeBody( byval s as FBSYMBOL ptr )
 
 		'' single-line comment?
 		case FB_TK_COMMENT, FB_TK_REM
-		    cComment( )
+			cComment( )
 
 		'' newline?
 		case FB_TK_EOL
@@ -805,9 +805,10 @@ private sub hTypeBody( byval s as FBSYMBOL ptr )
 			'' isn't it a field called TYPE|UNION?
 			select case as const lexGetLookAhead( 1 )
 			case FB_TK_EOL, FB_TK_EOF, FB_TK_COMMENT, FB_TK_REM, _
-			     FB_TK_FIELD
+				FB_TK_FIELD
 
-decl_inner:		'' it's an anonymous inner UDT
+decl_inner:
+				'' it's an anonymous inner UDT
 				isunion = lexGetToken( ) = FB_TK_UNION
 				if( symbGetUDTIsUnion( s ) = isunion ) then
 					errReport( FB_ERRMSG_SYNTAXERROR )
@@ -966,11 +967,11 @@ sub cTypeDecl( byval attrib as FB_SYMBATTRIB )
 
 	case FB_TKCLASS_QUIRKWD
 
-    case else
+	case else
 		errReport( FB_ERRMSG_EXPECTEDIDENTIFIER )
 		'' error recovery: fake an ID
 		checkid = FALSE
-    end select
+	end select
 
 	if( checkid ) then
 		'' Namespace identifier if it matches the current namespace
@@ -979,8 +980,8 @@ sub cTypeDecl( byval attrib as FB_SYMBATTRIB )
 		if( fbLangOptIsSet( FB_LANG_OPT_PERIODS ) ) then
 			'' if inside a namespace, symbols can't contain periods (.)'s
 			if( symbIsGlobalNamespc( ) = FALSE ) then
-  				if( lexGetPeriodPos( ) > 0 ) then
-  					errReport( FB_ERRMSG_CANTINCLUDEPERIODS )
+				if( lexGetPeriodPos( ) > 0 ) then
+					errReport( FB_ERRMSG_CANTINCLUDEPERIODS )
 				end if
 			end if
 		end if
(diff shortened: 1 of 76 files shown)


src/compiler/parser-expr-variable.bas
```diff
@@ -433,7 +434,7 @@ sub cUdtTypeMember _
 			'' ID
 			lexSkipToken( LEXCHECK_POST_SUFFIX )
 			dtype = symbGetFullType( sym )
-			subtype = symbGetSubType( sym )
+			subtype = sym
```
dkl, line 437: This doesn't seem right; it would only be correct if sym is always a struct, but that's not guaranteed here, or is it?

src/compiler/parser-decl-struct.bas
```diff
@@ -852,10 +853,19 @@ decl_inner:		'' it's an anonymous inner UDT
...
+			'' an inner type declaration and instead expect this is
+			'' going to be 'TYPE AS <datatype>'
+			case FB_TK_AS
 				hTypeElementDecl( FB_TK_DIM, s, attrib )
 
+			'' Otherwise TYPE|UNION starts an inner declaration
+			case else
+				'' allow named UNION|TYPE to be nested in another UNION|TYPE
+				hBeginNesting( s )
+				cTypeDecl( attrib )
```
dkl, line 867: So `cTypeDecl()` can be called recursively now. I noticed one `static` variable in it, `id`, that might need to be changed to normal to support that.

dkl (commented):
After looking at this a bit, it looks mostly sane, but I'm still wondering about some details.

One issue is: There doesn't seem to be a check to prevent a nested UDT to use the "parent" UDT, but this is needed because the parent UDT's definition isn't finished yet. For example: (currently no fbc error, but with -gen gcc there is an error due to the sizeof sanity check inserted by the backend)
```
type t1
	type t2
		as t1 x
	end type
	as byte x
end type

dim x1 as t1
dim x2 as t1.t2
print sizeof(x1), sizeof(x2)
```
If I remember correctly, FB doesn't allow recursive structs (obviously), but this adds a new dimension to that aspect. There are `hPatchByvalParamsToSelf` and `hPatchByvalResultToSelf`, but those only handle the one UDT, not any nested UDTs.

This recursion currently crashes fbc:
```
type t1
	type t2
		as t1 x
	end type
	as t2 x
end type

dim x1 as t1
dim x2 as t1.t2
print sizeof(x1), sizeof(x2)
```

src/compiler/symb-mangling.bas
```diff
@@ -741,8 +745,9 @@ private sub hMangleVariable( byval sym as FBSYMBOL ptr )
 	'' prefix
 	'' public global/static?
 	if( sym->attrib and (FB_SYMBATTRIB_PUBLIC or FB_SYMBATTRIB_EXTERN or _
-	                     FB_SYMBATTRIB_SHARED or FB_SYMBATTRIB_COMMON or _
-	                     FB_SYMBATTRIB_STATIC) ) then
+		FB_SYMBATTRIB_SHARED or FB_SYMBATTRIB_COMMON or _
```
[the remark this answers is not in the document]
jayrm: I am open to putting the visual alignment back in.  

I hand edited those lines and was contemplating a tab indent only strategy.  But `./contrib/tools/check-whitespace.bas` in repo should currently handle the leading spaces used for visual alignment now.   My underlying goal eventually is to use `./contrib/tools/check-whitespace.bas` on all fbc source commits and have it filter the white space to something standard-ish/desirable without much effort.

src/compiler/parser-expr-variable.bas
jayrm: Originally, with `subtype = symbGetSubType( sym )` I had this issue:
```
type T1
        __ as integer
        type T2
                N as integer
        end type
end type
print sizeof( T1.T2.N )

error 265: Symbol not a CLASS, ENUM, TYPE or UNION type, before 'N' in 'print sizeof( T1.T2.N )'
```
I thought what is happening here is that `cUdtTypeMember()` is only called after a failed attempt at `cTypeOrExpression()` which still passes back a dtype+subtype (struct+T1.T2).  Since the subtype is used as the parent in the call to `hMemberId()`, then getting the subtype of the subtype was skipping over a type level.  Checking for non-struct is at the top of the loop.

src/compiler/parser-decl-struct.bas
jayrm: Ah, indeed.  `id` is used after the call to `hTypeAdd()` so `id` will need to be on the stack.

src/compiler/parser-expr-variable.bas
dkl: Ah, I think I see the issue now: The dtype/subtype need to be updated as new context for the next iteration of that while loop, just as in `cUdtMember()`, to parse multiple "nesting" levels of `.` member accesses. I didn't know this was possible nowadays without using `dim`/`var` to define variables/instances, that's really nice.

Anyways, I've been trying to break it:
```
type T1
	innerfield as short
end type

type T2
	myfield as T1
end type

print sizeof( T2.myfield.innerfield )
```
result:
```
src/compiler/symb.bas(1429): assertion failed at SYMBLOOKUPAT: (ns->class =FB_SYMBCLASS_STRUCT) or (ns->class =FB_SYMBCLASS_NAMESPACE) or (ns->class =FB_SYMBCLASS_ENUM)
```
Looks to me like both situations are possible here - finding fields, methods, static variables as usual, but also the nested types now. So I think this code needs to check the symbol class in order to know how to update dtype/subtype. Basically treating types (structs) as a special case, and keep the old version for other cases.

jayrm: Thanks, your insight helps.  Agreed, existing code paths should stay the same.

I think I will also need to put something in `cUdtMember()` to error when we expect an expression but access the nested type (declaration) instead of a field:
~~~
type outer
	a as byte
	type inner
		a as short
	end type
end type

dim x as outer
print sizeof( x.inner.a )
~~~
I started adding tests to the test suite which will be included in the next commits.

src/compiler/parser-identifier.bas
```diff
@@ -506,7 +506,9 @@ function cParentId _
 		'' check visibility
 		if( parent <> NULL ) then
 			if( symbCheckAccess( sym ) = FALSE ) then
-				errReport( FB_ERRMSG_ILLEGALMEMBERACCESS )
+				if( (options and FB_IDOPT_ISDEFN) = 0 ) then
```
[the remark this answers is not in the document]
jayrm: This should be resolved in [669cffbb80423b927e7b24c70ab7a9d6c600ce44]

src/compiler/symb-mangling.bas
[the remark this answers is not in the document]
jayrm: For the files I've been working on, visual alignment is added back in [151f0f32fad971044fad79ef03fd4fc1cce0f190]

src/compiler/parser-decl-struct.bas
```diff
@@ -1057,6 +1068,23 @@ sub cTypeDecl( byval attrib as FB_SYMBATTRIB )
 			end if
 		end if
 
+		'' check that the base type is not a parent of the current type
+		if( baseDType = FB_DATATYPE_STRUCT ) then
+			'' is this an inner type? don't allow extending any undeclared parent UDT
+			dim context as FBSYMBOL ptr = any
```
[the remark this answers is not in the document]
jayrm: I had trouble working out the logic for a symbol lookup, so I added something similar: `symbIsParentNamespace(dtype,subtype,namespace=NULL)`.  

I think the logic makes sense and should only have to walk back through the parents of the symbol to perform the check.  The function checks if the symbol is the parent of the current (or specified) namespace..

Added in [d22042a754f09db2146bbc2966b2d460189fd772] and used in multiple places in the following commit.

src/compiler/parser-decl-struct.bas
```diff
@@ -1057,6 +1068,23 @@ sub cTypeDecl( byval attrib as FB_SYMBATTRIB )
...
+			context = symbGetCurrentNamespc( )
+			while( context <> @symbGetGlobalNamespc( ) )
+				if( symbGetClass( context ) = FB_SYMBCLASS_STRUCT ) then
+					if( context = baseSubtype ) then
+						baseSubtype = NULL
```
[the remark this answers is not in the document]
jayrm: This should be resolved in [669cffbb80423b927e7b24c70ab7a9d6c600ce44]

src/compiler/parser-expr-variable.bas
jayrm: Reverted to previous logic and added separate case in [c93f26676ba9f9a5aab7f7028476cb53de1a8512]

This should address the issues of the last couple of posts in this conversation.

src/compiler/parser-decl-struct.bas
jayrm: This should be resolved in [fa021f65ab0b620f7523e4da36e5f443fac6cdae]

jayrm:
> If I remember correctly, FB doesn't allow recursive structs (obviously), but this adds a new dimension to that aspect. There are `hPatchByvalParamsToSelf` and `hPatchByvalResultToSelf`, but those only handle the one UDT, not any nested UDTs.

In [4073c8b55bbe9c57f85aa23e8ff47053db032d34] I've added logic to recalculate sizes and real return types in `hPatchByvalParamsToSelf` and `hPatchByvalResultToSelf`, which seems to work for trivial types (a simple test added).  It seems a little aggressive iterating through every member multiple times, so maybe there is a possibility to optimize, either by waiting until we are at the outermost type, or caching / checking state information (in addition to what's already tracked) about what actually needs to be recalculated.

Non-trivial types will be tricky; to be honest not sure how many problems.  In theory a simple stack with outermost type pushed first and innermost types push last should manage the order.  But I'm not sure that even if we delay closing all the types/scopes until we've reached the end of the outermost type, we won't still have some possible circular dependency.  I don't know, it might be a prudent approach just to disallow certain inner/outer type relationships or uses.

jayrm:
I thought my next step was going to be rewrite how the type scopes were managed and closed, but in trying to break this new feature I notice I am running up against some already reported bugs unrelated to this change.
For example:
[#649 field initializers/member proc decls see inconsistent UDT state](https://sourceforge.net/p/fbc/bugs/649/)

PR #403, opened 2022-10-29
The default value of the parameter was passed incorrectly when calling the pointer function.

skyfish4tb:
When multiple function pointers of the same type are defined, their corresponding parameter initialization values are different, although their prototypes are the same. In this case, only the parameter initializers of the first function pointer will be used.

Type OptionalError
	Test1 As Sub(Byval value As Long=123)
	Test2 As Sub(Byval value As Long=456)
	Test3 As Sub(Byval value As Long=789)
End Type

Sub Test (Byval value As Long)
	print value
End Sub

Dim As OptionalError vOErr = ( @test, @test, @test )
vOErr.Test1()
vOErr.Test2()
vOErr.Test3()

'Output:
'       123
'       123
'       123

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -61,6 +61,7 @@ Version 1.10.0
 - sf.net #916: fbc: 'LINE INPUT [#1 | prompt,] variable [, maxchars]' requires length parameter for ZSTRING/WSTRING PTR variable
 - sf.net #916: fbc: Error on LINE INPUT [#1] if either missing or irrelevant max length argument
 - sf.net #892: fbc: Bug when default calling via a pointer to a 'Sub' defined with the only 'Any Ptr' optional parameter
+- github #403: fbc: The default value of the parameter was passed incorrectly when calling the pointer function (Skyfish) [optional parameter values are specified by the function pointer declaration]
 - sf.net #944: fbc: regression - local static arrays shouldn't have the array descriptor moved to global scope - partially revert #944 and add test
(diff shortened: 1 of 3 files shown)


jayrm:
This is a good change.  I don't mind adding some new tests to the test suite later.

```
function f1( byval arg as long ) as long
	return 1000+arg
end function 
function f2( byval arg as long = 2 ) as long
	return 2000+arg
end function 
function f3( byval arg as long = 3 ) as long
	return 3000+arg
end function 

'' default initializers are stored with the function pointer
var pf1 = procptr( f1 )
var pf2 = procptr( f2 )
var pf3 = procptr( f3 )

print "call functions by pointer"
? pf1(1)  '' 1001  <-- f1(1)
? pf2()   '' 2002  <-- f2(2)
? pf3()   '' 3003  <-- f3(3)

'' swap proc addresses, but not initializers
'' because function pointer declaration is still
'' same as originally typed

print "call functions by pointer (swap 2+3)"
swap pf2, pf3

? pf1(1)  '' 1001  <-- f1(1)
? pf2()   '' 3002  <-- f3(2)
? pf3()   '' 2003  <-- f2(3)

print "call functions by pointer (swap 1+2)"
swap pf1, pf2

? pf1(1)  '' 3001  <-- f3(1)
? pf2()   '' 1002  <-- f1(2)
? pf3()   '' 2003  <-- f2(3)
```

PR #411, opened 2023-06-13
Added headers for FLTK library

angros47:


jayrm:
Thanks for making a pull request.

What version of FLTK are these headers based on?

Should `./inc/FLTK/abi-version.bi ` also be included?  Checks on `FLTK_ABI_VERSION` in `Fl_Text_Display.bi` and `Fl_Window.bi` are commented out.

What is the origin of `__FB_APPLE__`?

angros47:
I converted the headers from the version 1.3.5, and that is the version I ran all my tests on Linux.

The ABI exposed by that library, by the way, seems to be by default the same of version 1.3.0 (in fact my headers work even with the DOS port from XFDOS, that is based on FLTK 1.3.0)

in the C++ headers the file abi-version.h I had was empty, it contained only comments that stated "DO NOT EDIT - This file is generated by configure !", so there was no need to port it: it would be useful for libraries compiled with settings different from the default ones, so it would make sense only if we modified the building process of FLTK to provide custom version of the FreeBasic headers. From what I figured out from my tests, by default the FLTK library provides the ABI for version 1.3.0, and this is what I have targeted.

About __FB_APPLE__, I put it as placeholder: as far as I know, there is no official version of FreeBasic for Apple computers, only some unofficial ports, but I wouldn't be surprised there will be an official port in future. If that will happen, I expect that version to have an intrinsic define, like the DOS, the Windows and the Linux versions. The code in the blocks marked with __FB_APPLE__ is never compiled, and I haven't tested it. I left it as reminder that, if there will ever be a version for Apple computer, that part of the header has to be adjusted.

rversteegen:
I'm surprised to see headers for such a big C++ library; I'm impressed if FB's C++ ABI compatibility is now good enough to support this.

What tool did you use to generate the headers? If it's fbfrog, can you provide the fbfrog configuration? (Probably) most headers included with FB are generated by fbfrog in the [fbbindings](https://github.com/freebasic/fbbindings) repo.

Because this is FLTK 1.3.5, and FLTK 1.4 has already existed for many years but isn't released yet, how would the different FLTK versions be distinguished? Would future headers go in `inc/FLTK1.4/` since these are `inc/FLTK`?

You're mistaken, FB does have an official port for Mac, but there are no official binaries and it's incomplete (libfbgfx isn't ported). The intrinsic define is `__FB_DARWIN__`, which is very unintuitive. Something like `__FB_MAC__` or `__FB_APPLE__` as an alias may be a good idea.

angros47:
I used no tool for such conversion, I did it by hand, in the hard way, unfortunately (that's why it took me so much time)

And I have the impression FLTK1.4, if they will ever be done, should indeed go in a different directory, since it seems to be treated, even on the FLTK site, as a different library.

jayrm:
> From what I figured out from my tests, by default the FLTK library provides the ABI for version 1.3.0, and this is what I have targeted.

https://github.com/fltk/fltk/blob/branch-1.3/README.abi-version.txt

If targeting 1.3.0 (i.e. all the testing, etc) then I feel like the guards in the headers should be preserved (not simply commented out).  For example something like:
`#if defined(FLTK_ABI_VERSION) andalso (FLTK_ABI_VERSION >= 10303)`

Or maybe better in a common header:
```
#ifndef FLTK_ABI_VERSION
#define FLTK_ABI_VERSION 10300
#endif
```
1) Default is 1.3.0
2) Newer ABI breaking changes can still optionally be included in the translated headers
3) User can override to newer ABI by defining FLKT_ABI_VERSION themselves in their own source

Or conversely. target a later 1.3.4 version for testing and development:
```
#ifndef FLTK_ABI_VERSION
#define FLTK_ABI_VERSION 10304
#endif
```
1) Default is 1.3.4
2) User can override to older ABI by defining FLKT_ABI_VERSION themselves in their own source

angros47:
I would suggest to target 1.3.0, since it is the default version, and also, the DOS version is only 1.3.0, and I think the ability to build GUI application for DOS, too, is an useful feature.

Adding support for 1.3.4 would require redoing almost every test on all platforms, so it's not a trivial task

jayrm:
I see this pull request is same as source located at https://github.com/angros47/FLTK-headers-for-FreeBasic

An alternate way to manage this contribution would be to have FLTK-headers-for-FreeBasic be a sub module of fbc.  Using a sub-module has some benefits for on-going development but then also requires extra steps for fetching fbc and building releases so there are trade-offs.

I don't really have time to look at usage of these headers more at this time, so simplest seems to just go for it and merge it in, and see what kind of user feedback we get.  I will guess users will ask things like: where do I get the FLTK libraries? how to compile FLTK programs? where are the example programs?  Why does my program crash with version x.y.z? etc...

angros47:
I am currently trying to port some examples to FreeBasic, on the forum

PR #425, opened 2024-04-14
Initialize variables

zamabuvaraeu:
Initialize variables with NULL and 0 value because it may be used uninitialized


jayrm:
Thank-you.  What build version of tool was being used when finding these?  I suspect these commits were to fix warnings in gcc / clang.

zamabuvaraeu:
I do not have any special tools.

I am compiling FreeBASIC with the `-gen gcc -O 3` options. GCC issues a lot of warnings about uninitialized variables.

I selected only those warnings where it is clear that variables can be used in an uninitialized form. Here is a typical example of such code:

```freebasic
Function Foo() As Any Ptr
	Dim vr As Integer Ptr = Any

	If ( condition ) Then
		vr = ...
	End If

	' if the condition is not met, the variable will remain uninitialized
	function = vr
End Function
```

PR #426, opened 2024-05-31
Preprocess the target macro based on the macro parameter list.

skyfish4tb:
'' __FB_ARG_LISTEXPAND__(MacroName, MacroArgCount , args...)
	'' Retuns empty string on invalid index, rather than compile error.
	'' MacroName     : the name of the macro used for extension calls.
	'' MacroArgCount : the number of parameters for the MacroName macro.Expand the parameter list Args according to the MacroArgCount value.
	''                 1, MacroArgCount>0, pass MacroArgCount parameters each time;
	'                  2, MacroArgCount=0, pass all parameters;
	''                 3, MacroArgCount<0, for each parameter passed, the previous MacroArgCount parameters are automatically removed on the next pass.
	'' args...       : argument list(The number of Args parameters passed may not be an integer multiple of MacroArgCount.)

	#macro m( arg... )
	    #print "   "##arg
	#endmacro

	#print "1. MacroArgCount=0:"
	__FB_ARG_LISTEXPAND__( m, 0, Hello1, Hello2, Hello3, Hello4)
	#print "2. MacroArgCount>0:"
	__FB_ARG_LISTEXPAND__( m, 1, Hello1, Hello2, Hello3, Hello4)
	#print "3. MacroArgCount<0:"
	__FB_ARG_LISTEXPAND__( m, -1, Hello1, Hello2, Hello3, Hello4)

	/' Compiler output:
	   1. MacroArgCount=0:
	      Hello1, Hello2, Hello3, Hello4
	   2. MacroArgCount>0:
	      Hello1
	      Hello2
	      Hello3
	      Hello4
	   3. MacroArgCount<0:
	      Hello1, Hello2, Hello3, Hello4
	      Hello2, Hello3, Hello4
	      Hello3, Hello4
	      Hello4
	'/


jayrm:
This appears to be an improvement over the commit in #312 `__FB_ARG_LOOPEXPAND__`

While it does seem of specific use, this new version generally looks OK to add.

It needs a couple of corrections:
1) `sizeof(wstring)` may vary on different hosts so `ushort ptr` can't be assumed.
2) NEWLINE should only be added if the macro expansion results in multiple lines, otherwise it throws off the internal line number counter
3) tests in the test-suite

I will post fixes soon.

PR #436, opened 2025-01-28
doc: Convert to pcre2 from obsolete pcre (pcre3)

thekhalifa:
libpcre (a.k.a pcre3) is obsolete in favour of pcre2 by upstream. While pcre is still available, it is being removed from several linux distributions [1] [2]

- Change CRegex to use the new version and remove inclusion of the old pcre library.
- Remove inclib "pcre" from fbchkdoc. Note: not replaced with pcre2 here as this should be included from the pcre2.bi header

Tested on linux with only libpcre2 installed

1. https://fedoraproject.org/wiki/Changes/PcreDeprecation
2. https://lists.debian.org/debian-devel/2025/01/msg00621.html


doc/libfbdoc/CRegex.bas
```diff
@@ -88,27 +92,26 @@ namespace fb
...
+		if( ctx->matchdata <> NULL ) then
+			pcre2_match_data_free(ctx->matchdata)
+			ctx->matchdata = NULL
+		end if
+		if( ctx->code <> NULL ) then
+			pcre2_code_free(ctx->code)
+			ctx->code = NULL
+		end if
+		if( ctx->ccontext <> NULL ) then
+			pcre2_compile_context_create( ctx->ccontext )
```
jayrm, line 111: Shouldn't this free the compile context?
`pcre2_compile_context_free( ctx->ccontext )`

jayrm:
I am feeling impatient today, I will make the edit and update the changelog.txt.  Thanks for the PR.

thekhalifa:
Well spotted, yes, that was definitely meant to _free() not _create(). And thanks.

PR #438, opened 2025-01-31
Add examples makefile

thekhalifa:
Makefile to compile all examples/ along with minor compile fixes to examples/compression

When reading the examples, it's helpful to have them compiled already to compare output.
Also various examples are slightly out of date with modern libraries and FreeBASIC, so attempting to correct some of those examples along the way.


jayrm:
This is a good idea and a good starting point.  I will make a few edits for windows (if this PR allows me) and then merge.

PR #440, opened 2025-02-22
examples: Add portaudio sine wave example

thekhalifa:
PortAudio already has a FB header, this adds an example usage to play a sine wave from a buffer.

### DIFF
--- a/examples/sound/portaudio-sine.bas
+++ b/examples/sound/portaudio-sine.bas
@@ -13,7 +13,7 @@ Type PlayPayload
     frequency As Single
 End Type
 
-Function PlayCallback (InputBuff as Const Any Ptr, OutputBuff as Any Ptr, _
+Function PlayCallback cdecl (InputBuff as Const Any Ptr, OutputBuff as Any Ptr, _
                     FrameCount as CUlong, TimeInfo as Const PaStreamCallbackTimeInfo Ptr, _
                     StatusFlags as PaStreamCallbackFlags, UserData as Any ptr) as Long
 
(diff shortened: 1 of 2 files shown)


examples/sound/portaudio-sine.bas
```diff
@@ -0,0 +1,82 @@
...
+
+Const Pi as Double = 3.14159265358979323846
+Const SampleRate as Integer = 44100
+Const SineSamples as Integer = SampleRate / 10
+
+Type PlayPayload
+    frequency As Single
+End Type
+
+Function PlayCallback (InputBuff as Const Any Ptr, OutputBuff as Any Ptr, _
```
jayrm, line 16: On windows, the calling the convention will default to `stdcall` but the `portaudio.bi` API expects `cdecl`
This callback function should be defined with `cdecl` calling convention as in `Function PlayCallback cdecl ( ...` for portability to windows.

examples/sound/portaudio-sine.bas
```diff
@@ -0,0 +1,82 @@
...
+Sub Play(Frequency as Single, Duration as Single)
+    Dim as PaError pe
+    Dim as PaStream Ptr stream
+
+    Dim as PlayPayload payload
+    payload.frequency = Frequency
+
+    pe = Pa_OpenDefaultStream(@stream, 0, 1, paInt32, SampleRate, SineSamples, @PlayCallback, @payload)
+    If pe <> paNoError Then
+        Print "Error: "; Pa_GetErrorText(pe)
```
jayrm, line 38: `Pa_GetErrorText(pe)` here and a few other locations below are missing the dereference `*` to print the pointed to string.  i.e. `*Pa_GetErrorText(pe)`

jayrm:
I tried this example on windows after hacking together a static library compiled from sources from https://github.com/PortAudio/portaudio 

This example worked for me on windows 32 bit.  I didn't try 64-bit.  Only issue I had with the example itself is that it worked on 48000Hz samplerate, but not 44100Hz.  Don't know if that issue is windows thing or a hardware thing or I missed something building the library.

thekhalifa:
> I tried this example on windows after hacking together a static library compiled from sources from https://github.com/PortAudio/portaudio
> 
> This example worked for me on windows 32 bit. I didn't try 64-bit. Only issue I had with the example itself is that it worked on 48000Hz samplerate, but not 44100Hz. Don't know if that issue is windows thing or a hardware thing or I missed something building the library.

Fished out a Win32 vm and I think it was a latency issue where the stream was stopped before it started playing. The callback wasn't called at all when testing portaudio with the Win MM API (I couldn't compile it with WASAPI or DX).

So now it's 3 notes that play for longer and prints out the API name. Sounds a bit clippy on Windows, but should be good enough as an API example. The original sequence was meant as the game over sound, but the buffer playback needs to be more clever to keep the stream open.

jayrm:
Thank-you for the fix ups and the extra testing.  I agree is good enough for a start point and should not be too difficult for users to play test on various windows versions and hardware to find what works with error messages shown.  Thanks.

PR #442, opened 2025-02-23
libfbdoc SSL: try to use native CA certs, fix readme

swurzinger:
this PR contains two small changes:
- **fix fbchkdoc documentation mentioning wrong argument names for CA certs**
  argument names mentioned in the readme did not match the ones in https://github.com/freebasic/fbc/blob/master/doc/fbchkdoc/cmd_opts.bas#L577
- **enable native CA by default if no custom CA file specified**
  this will make it work on windows and a few other supported systems without specifying a CA file. see https://curl.se/libcurl/c/CURLOPT_SSL_OPTIONS.html for details  
  I tested it on Windows 11 and Ubuntu 20.

### DIFF
--- a/doc/fbchkdoc/readme.txt
+++ b/doc/fbchkdoc/readme.txt
@@ -175,12 +175,12 @@ the '-ini FILE' command line option to any of the tools.
         are given.  Set or override this url with the '-url URL' command
         line option
 
-    web_cert_file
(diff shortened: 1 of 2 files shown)


jayrm:
Thank-you for this fix.  OK for me on tests.

PR #448, opened 2025-03-31
Add support for Cocoa OpenGL screen in gfxlib2 and other compiler error fixes for macOS 

Markos-Th09:
This PR includes several changes to improve support for the macOS platform in fbc and its libraries. It also aims to add support for OpenGL screens using Cocoa instead of requiring XQuartz for X11 support, although at the time of writing this I haven't tested the implementation very thoroughly yet.

Currently some mechanism similar to `#inclib` but for macOS frameworks is missing. Since a lot of important system libraries contained within frameworks such as OpenGL, this would be useful for headers such as `gl.bi` and `glut.bi` but also in general.


rversteegen:
Wow, nice!

I'm not sure about frameworks, but I think it should be up to the user whether they want to link to a framework or to a dylib. For example you can either install SDL via homebrew or compile/download a framework. For system frameworks there's no need for a choice, Maybe a new directive to add a framework, with some way to disable it (do I remember that there's now a way to disable/remove libraries added by #inclib?)

I'll want to look into the computed goto removal. clang does support computed gotos. Casting a label address to a function pointer and calling it seems very unsafe (or rather, x86-specific).

Markos-Th09:
Just to be sure when using gfxlib functions like `Line` or `Circle` am I supposed to see nothing in an OpenGL `Screen`?  It seems to also be the case on X11. What tests do I need to perform to make sure the implementation is fully correct?

PR #470, opened 2026-05-15
Fix macOS arm64 (Apple Silicon) linker support

metaneutrons:
## Summary

Fix linking on macOS arm64 (Apple Silicon, M1/M2/M3/M4).

## Changes (3 files, +47/-11)

**makefile:**
- Normalize `arm64` (from `uname -m`) to `aarch64` before the 32-bit `armv%` catch-all

**src/compiler/fbc.bas:**
- `line input` in `hGet1stOutputLineFromCommand` (handles spaces in SDK paths)
- Add `-arch arm64` for `FB_CPUFAMILY_AARCH64` on darwin
- Replace deprecated `-macosx_version_min` with `-platform_version macos <min> <sdk>`
  - min: 11.0 for arm64, 10.4 for x86_64 (preserves original minimum)
  - sdk: detected via `xcrun --show-sdk-version` (fallback 14.0)
- SDK path via `xcrun --show-sdk-path` (only on darwin host, respects user `-sysroot`, quoted with QUOTE)
- Skip GNU `--sysroot=` on darwin (Apple ld uses `-syslibroot`)
- Remove `--eh-frame-hdr` on darwin (unsupported by Apple ld)
- Remove `-lgcc` from darwin default libs (unavailable with Apple clang)

**src/rtlib/profile_cycles.c:**
- Guard ELF `__start_`/`__stop_` section boundary symbol declarations and usage with `!defined(HOST_DARWIN)`
- Provide `data=NULL, length=0` fallback on darwin (cycle profiling compiles but produces no output)

## Testing

```
make bootstrap-minimal ENABLE_STANDALONE=1
```
Successfully builds fbc and runtime library on macOS 15 (Sequoia) with Apple clang 21 on M4.

## Notes

- x86_64 darwin builds unaffected (minimum version preserved at 10.4)
- `-profile cycles` compiles on darwin but produces no per-procedure data (Mach-O lacks ELF section boundary symbols)
- Bootstrap requires `bootstrap/darwin-aarch64/` (copy from `bootstrap/linux-aarch64/` + fix clang computed goto)

### DIFF
--- a/makefile
+++ b/makefile
@@ -342,8 +342,13 @@ ifneq ($(filter 386 486 586 686 i386 i486 i586 i686,$(TARGET_ARCH)),)
   TARGET_ARCH := x86
 endif
 
-# Normalize TARGET_ARCH to arm
-ifneq ($(filter arm%,$(TARGET_ARCH)),)
+# Normalize TARGET_ARCH to aarch64 (macOS reports arm64)
+ifneq ($(filter aarch64 arm64,$(TARGET_ARCH)),)
+  TARGET_ARCH := aarch64
+endif
+
+# Normalize TARGET_ARCH to arm (32-bit only, after aarch64 is handled)
+ifneq ($(filter armv% arm,$(TARGET_ARCH)),)
   TARGET_ARCH := arm
 endif
 
--- a/src/compiler/fbc.bas
+++ b/src/compiler/fbc.bas
@@ -305,7 +305,7 @@ private function hGet1stOutputLineFromCommand( byref cmd as string ) as string
 	end if
 
 	dim ln as string
-	input #f, ln
+	line input #f, ln
 
 	close f
 	return ln
@@ -840,6 +840,8 @@ private function hLinkFiles( ) as integer
 		case FB_CPUFAMILY_ARM
 			'' fixme: this is clearly too specific
 			ldcline += "-arch armv6 "
+		case FB_CPUFAMILY_AARCH64
+			ldcline += "-arch arm64 "
 		end select
 	end select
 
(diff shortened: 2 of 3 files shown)


agorangetek:
Nice work on getting the platform pieces in order — the `arm64` → `aarch64`
normalization here is exactly right; without it a plain `make` on Apple Silicon
resolves to `darwin-arm`, and I've borrowed that change for #479 with credit.

While comparing our two approaches I ran into two things that stop this branch
from producing a working link, in case they are useful:

**1. `crt1.o` cannot be found on current macOS.** `hLinkFiles()` adds it
explicitly for Darwin executables through `hFindLib("crt1.o")`:

```
$ clang --sysroot=$(xcrun --show-sdk-path) -print-file-name=crt1.o
crt1.o          # bare name => unresolved, even though the SDK contains the file
```

`fbcBuildPathToLibFile()` treats an unresolved name as failure
(`hStripPath(found) == found`), so fbc reports

```
error 23: File not found, crt1.o
```

and no startup object reaches the link line. ld64 then fails with:

```
ld: Undefined symbols for architecture arm64:
  "_main", referenced from: <initial-undefines>
```

because the SDK's `crt1.o` is what provides the entry point that calls `main`.
Letting clang be the link driver (or simply skipping `crt1.o` on Darwin) is what
resolved it for me.

**2. arm64 needs the DATA descriptor aligned.** It is emitted as a packed
`{ short, void* }` table, so the embedded pointers sit at 2-byte offsets and
ld64 rejects the relocation outright, regardless of how the link line is built:

```
ld: pointer not aligned in '_label$N'+0x16
```

That one is a hard error on arm64 (it is only a warning on x86_64).

#479 fixes both, plus a couple of other arm64-specific items: `va_list` is a
plain pointer on Apple arm64 rather than the AAPCS64 `__va_list_tag` struct, and
the X11 `XGETKEYBOARDMAPPING` typedef no longer matches modern Xlib when
compiled with clang. It also passes the fbcunit suite and handles `-dylib`.

Not trying to compete with this PR — happy for whichever lands first to win; I
just did not want the `crt1.o` part to be a surprise, since it is the first
thing that fails.

metaneutrons:
Closing this outdated Apple Silicon patch series. Its fixes have been separated into independently testable upstream contributions; the unrelated tool-output fix is now #482 and will not be folded into this PR.

